PR #1, opened 2014-07-11
Harmonic static site generator

jaydson:
Adding Harmonic static site generator to the list.


jaydson:
ping @sindresorhus 
Almost one year later, can you check it out the current state of Harmonic now?
https://github.com/JSRocksHQ/harmonic
The project is pretty active and soon we'll release a dedicated website.
The current stable version is 0.4.2.

Thanks.

PR #228, opened 2015-01-05
learnyounode course, few algorithm implementations

tejasmanohar:
I added learnyounode course as well as a CS section with some CS algorithms implemented as Node.js modules (more to come!).


sindresorhus:
I already link to Nodeschool which as more than that course. Not interested in the algo modules as they're not very commonly useful.

tejasmanohar:
Ok. Makes sense :+1:

PR #229, opened 2015-01-08
add instrumentation category and strong-agent entry

altsang:
Link to package - https://github.com/strongloop/strong-agent

Should include into the awesome list because this is a full package solution for application monitoring NOT locked into a dashboard like New Relic.  With the StatsD interface you're free to take your application metrics to any supported dashboard provider or solution including - Graphite, Datadog, Zabbix, etc.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -13,6 +13,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [HTTP](#http)
 - [Debugging / Profiling](#debugging--profiling)
 - [Logging](#logging)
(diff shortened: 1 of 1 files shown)


sindresorhus:
No thanks:

> NOTE: This project is not released under an open source license. We welcome bug reports as we continue development, but we are not accepting pull requests at this time.

altsang:
got it - would be good to clarify in contribution.md that you're only looking for OSS licensed entries and maybe list which kinds of OSS licenses are acceptable , e.g. MIT, Apache2, Artistic, AGPL(???)

PR #230, opened 2015-01-08
Add Grant to the Authentication section

simov:
OAuth middleware for Express and Koa - the simplest way to deal with OAuth in the NodeJS world. You can take a look at the code and the documentation https://github.com/simov/grant and the showcase https://grant-oauth.herokuapp.com/

Currently supported providers **84** with Hapi support underway.


simov:
Thanks :+1:

PR #232, opened 2015-01-09
added strong-log-transformer to logging section

altsang:
Link to package - https://github.com/strongloop/strong-log-transformer

I think this should be added to the awesome list because it's a lightweight to perform the operations for the most common log format and mask operations.


sindresorhus:

PR #236, opened 2015-01-23
Add Ramda as Functional Programming Library

geekflyer:
Ramda grew into a serious, popular Functional Programming Library in recent months. Got already ~1500 stars on Github: https://github.com/ramda/ramda

It doesn't have as many built-in functions like Lo-Dash, but the way it enables functional composition is often more concise.
For example instead of chaining one usually uses R.compose or R.pipe and then plugs in different Ramda functions (which are automatically curried) OR user-built functions. Ramda functions and user-built functions are treated equally, as oppose to chaining.


sindresorhus:
Yeah, looks interesting.

PR #238, opened 2015-02-01
Add Unit.js

Nicolab:
Unit.js is an assertion library, running on Node.js and the browser. It works with any test runner and unit testing framework like Mocha, Jasmine, Karma, ... 
Supports dependency injection and is extensible via a plugins system. 

Fluent style :
```js
test.array(['a', 'b', 'c']).contains(['a','c']).hasLength(3);
```

Example, tests of [an event emitter](https://github.com/Nicolab/evemit/blob/master/test/test.js).

Fully tested and documented.


sindresorhus:

PR #239, opened 2015-02-01
add if-async module

kessler:
conditional async execution for async.js


sindresorhus:

PR #240, opened 2015-02-03
Added KrakenJS to web frameworks.

walmik:
KrakenJS is extends Express in interesting ways providing security, npm proxy as well as helps in Globalization by making provision for Internationalization and Localization right out of the box. It also adds common grunt tasks for Less, Mocha, functional testing with Nemo and much more.


sindresorhus:
Duplicate of #155

PR #241, opened 2015-02-03
Add markup.js

alexbeletsky:
Compact and powerful templating library. 

I use that both front-end and backend, very happy with simplicity and features. It's mentioned on [template-chooser](http://garann.github.io/template-chooser/).

https://github.com/adammark/Markup.js


sindresorhus:

PR #242, opened 2015-02-03
Add rest.js to http section

jenius:
Rest is a fantastic promise-based http client, super clear and robust, with a number of built-in modules (called interceptors) to help configure defaults for your requests.


sindresorhus:

jenius:
Just out of curiosity, how come?

PR #243, opened 2015-02-19
Added MEAN.JS framework

starquake:


sindresorhus:
Thanks, but not really a Node module.

PR #245, opened 2015-02-25
Add command-promise

StreetStrider:
This is my own lib which I use in deploying and automation to run shell scripts. In general: it makes `child_process.exec` return promise. It has some additional features:
* It flattens arrays and arguments list passed, so no need to `join` arguments or do `apply`. This allows to pass some hardcoded data (like command name, e.g. `ls`) and variable data (e.g. path, or array of paths).
* It has utils for transforming command output.
* It has simple form for inserting in promise flow and also `Command` can be `partial`led or `bind`ed and not depends on `this`.

I tried to notice some of features in description, but description should be not only descriptive, but also short, so I just put «promise wrapper», as I say above, this is not the only feature.
I'm not sure about section for this, but minimist here, so I put this stuff here too.

I was using this lib for a long time and it has good pack of tests.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -181,6 +181,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [cli-cursor](https://github.com/sindresorhus/cli-cursor) - Toggle the CLI cursor.
 - [ascii-charts](https://github.com/jstrace/chart) - ASCII bar chart in the terminal.
 - [cfonts](https://github.com/dominikwilkowski/cfonts) - Sexy ASCII fonts for the console.
+- [Command](https://github.com/StreetStrider/command-promise) - Promise wrapper for child_process.exec.
 
 
 ### Package managers


sindresorhus:
Thanks, but not a big fan of promise wrappers tbh.

Also this should definitely be an external dependency: https://github.com/StreetStrider/command-promise/blob/master/src/deps.js

StreetStrider:
@sindresorhus, hi again. You're right about `deps.js` it can completely solved by our awesome npm/node dependency system. But I've done this for perfomance. Every time I bundle my deps in some other projects this is resulted in reduced startup time. Which is great for those cases. And I was thinking this would be great for things like working with shell commands too. The only side-effect is that deps are fixed at certain version (and will be fixed until I come again and update them by hand). Anyway, you're right, this is kind of redundant thing.

> promise wrappers

This was started as promise wrapper, but as I said before, it has a bunch of pretty features. By calling it only a «promise wrapper» you're making me sad :)

StreetStrider:
@sindresorhus, hi. Can you take a look at this again? I've added one another interesting interface, which converts child processes into duplex streams. Very neat in use. I've also turned off bundling deps by default, preserving it as a npm task which can be called deliberately after installing.

PR #248, opened 2015-03-12
Add validator.js to Data validation section

aspyatkin:
There is only one library in Data validation section, so I has added one more.


sindresorhus:

aspyatkin:
Something wrong with PR?

PR #251, opened 2015-03-16
Add Diet.js to Web frameworks section

mr-taviscaron:


sindresorhus:

PR #260, opened 2015-04-13
Add Osmosis to Parsing

lpinca:
There are no web scrapers in the list so I thought it would make sense to add one.

[Osmosis](https://github.com/rc0x03/node-osmosis) is an HTML/XML parser and web scraper with the following features:

- Fast: uses libxml C bindings
- Lightweight: no dependencies like jQuery, cheerio, or jsdom
- Clean: promise based interface- no more nested callbacks
- Flexible: supports both CSS and XPath selectors
- Predictable: same input, same output, same order
- Detailed logging for every step
- Easy debugging with built-in stack size and memory usage reporting

Maybe it makes sense to also add [x-ray](https://github.com/lapwinglabs/x-ray).


sindresorhus:
I'd prefer to only add `x-ray`. Osmosis isn't very nodey.

PR #261, opened 2015-04-13
added iwebpp.io in Mad science

sequoiar:
iWebPP.io intends to enable peer and p2p style web service behind nat/fw


sindresorhus:

PR #263, opened 2015-04-16
Add emailjs - a suite for IMAP/SMTP and email parsing

flxw:


sindresorhus:

PR #264, opened 2015-04-20
Add node_pcap

John-Lin:
The `node_pcap` package is a awesome library for raw packet capture (Both live capturing and pcap file parsing), decoding, and analysis. It is a set of bindings from `libpcap` which is packet capture library used by programs like `tcpdump` or `wireshark`.

Here is the repo https://github.com/mranney/node_pcap and published package on NPM https://www.npmjs.com/package/pcap.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -378,6 +378,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 
 ### Network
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Yeah, looks nice, but I'm trying to collect things that are more generally useful. This is more of a specialized tool.

PR #265, opened 2015-04-20
Add is.js to Data validation section

aspyatkin:
is.js is an excellent library to check objects - including type checks, arithmetic checks, string checks and a lot of others. I think it's worth adding it to this awesome list!


sindresorhus:
I'm sure it's useful, but not very nodey and too unfocused in my opinion.

PR #267, opened 2015-04-23
Add `navit`, remove `Nightmare`

puzrin:
[navit](https://github.com/nodeca/navit) is  similar to [Nightmare](https://github.com/segmentio/nightmare) from your list, but has much more consistent API (including intermediate errors handle) and better phantomjs bridge. Don't know is it strong enougth for this list. But i'm sure you will see a big difference if start to use.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -455,7 +455,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [istanbul](https://github.com/gotwarlost/istanbul) - A code coverage tool that computes statement, line, function and branch coverage with module loader hooks to transparently add coverage when running tests.
 - [Sinon.JS](https://github.com/cjohansen/Sinon.JS) - Test spies, stubs and mocks.
 - [Karma](http://karma-runner.github.io) - Executes code in multiple real browsers.
-- [Nightmare](http://nightmarejs.org) - High-level PhantomJS wrapper that lets you automate browser tasks.
+- [navit](https://github.com/nodeca/navit) - PhantomJS wrapper for easy client tests scripting.
 
 
 ### Benchmarking


sindresorhus:
More consistent how? How is the phantomjs bridge better? (this should really be somewhere in your readme, as I as a user would probably like to know how it compares to the major other competitor) ;)

puzrin:
I can reply to your direct questions, but don't like to place it in readme, because it will look like abuse of nightmare authors. If you know how to do it polite - let me know.

1. https://github.com/segmentio/nightmare/labels/performance - this issues are caused by bad driver.  Those are not fixed for a long time. I don't know why.
2. Nightmare's stack of commands does NOT interrupts on errors (you can throw exception, but can't interrupt more gently). For example, if you call .wait(selector) with timeout, you can't check if wait succeeded of timeout happened. That makes error checks much more complicated.
3. Sometime it's not clear if method is getter or setter (by name)
4. Passing additional params to functions (like eval) is almost useless because you can't make them depend on previous steps. In `navit` we have lazy evaluation for that - you can pass fns instead of direct values.
5. Too few assertions. It looks like nightmare was done to do actions, but not to validate results. See available tests in navit, from here and below https://github.com/nodeca/navit/blob/master/lib/test.js#L71 . You can check almost everything in one line. In nightmage that will be get + closure, more code.

That's my personal opinion. We used nightmare in nodeca. Then i've posted summary to their tracker and didn't noticed interest. After some time i've decided that it will be more simple to reinvent one more wheel :)

puzrin:
There is 2 more packages of similar style:

- https://github.com/johntitus/node-horseman, but autor's decided to make sync api, and seriously limited chainability (you can't do multiple tests in simple manner).
- https://github.com/assaf/zombie - they use jsdom instead of phantom/slimer, and this idea is too strange for me.

In navit i've tried to do ~ casper's api, but in curried style. The main reason why we don't use casper - because it requires separate enviroment and does not allow to interact with server & browser from single script (for example, load fixtures to db, then open page and do click)

sindresorhus:
> because it will look like abuse of nightmare authors

I don't see how that's abuse, you're only detailing why you decided to make something better. Most users would probably choose Nightmare since it's more popular while being ignorant to its shortcomings. It's in your best interest to explain to the user how your solution is better.

-

I'm convinced though, but I try not to have multiple things of the same thing. Can you remove Nightmare in favor of yours?

puzrin:
> Can you remove Nightmare in favor of yours?

Done.

PR #279, opened 2015-05-16
Update readme.md

LambertEliott:
Adding favicons + node-identicon + fav-identicons links to image part.


sindresorhus:
:palm_tree: Thanks for the suggestion, but I'm going to pass on these :)

PR #281, opened 2015-05-19
add EJS

hetz:
EJS is template module, very popular.


hetz:
:ok_hand:

kevva:
@hetz, why did you close this?

PR #287, opened 2015-05-28
New "Remote calls and distributed systems" section with Futoin packages

andvgal:
**Note: this PR depends on the previous #286**

This one does not follow the contributing guidelines - two packages are added with FutoIn sub-section pointing to intro. Perhaps, more packages will be added over time.

A new section for "Remote calls and distributed systems"

FutoIn - is a generic set of implementation-neutral enterprise grade remote call interface specifications and its reference implementations:

FutoIn Invoker (NPM: futoin-invoker):  https://github.com/futoin/core-js-ri-invoker
FutoIn Executor (NPM: futoin-executor): https://github.com/futoin/core-js-ri-executor

The overall idea of those two packages with corresponding specifications:
* define a neutral message coding format
   * JSON the default
   * Optional HTTP REST-like for human-made calls and large data uploads
   * TBD. Binary message coding equal to JSON in structure.
* define a neutral interface definition format
   * basic parameters types: boolean, integer, number, string, array and map
   * custom parameters: derived from the basic types
   * interface inheritance - in case if only little extensions are required and the functionality still can be used through neutral interface
   * interface import/mixin - define complex types once
   * strict interface versioning and backward compatibility requirements
* provide a concept of generic components for all-in-one or distributed-over-many-nodes project with no difference to invoking code.
   * connection and credentials manager (Invoker CCM) provides dependency injection feature in fact.
   * unified implementation for error-prone input message and parameter validation
* add out-of-the box advanced parameter type and checking
* support uni- and bi-directional communication channels
* support "blocking" and non-blocking calls (no response expected)
* support service-side callbacks
* support transport and message auditing
* integrate message security:
    * optional [HMAC](http://en.wikipedia.org/wiki/Hash-based_message_authentication_code) for message integrity checking
    * optional interface constraints like SecureChannel (requires secure network transport)
* extendable authentication types
* various fine tune options (e.g. timeouts), transport channel state handling, user-defined transport channel context/state variables etc.
* still make it possible to implement not fully complaint client or server in any technology in just a few minutes. It is even possible to make calls with command line cURL/wget using HTTP REST-like integration.
* fix uncertainty of REST and complexity of SOAP.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -49,6 +49,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [Process management](#process-management)
 - [AST](#ast)
 - [Weird](#weird)
+- [Remote calls and distributed systems](#remote-calls-and-distributed-systems)
 - [Miscellaneous](#miscellaneous)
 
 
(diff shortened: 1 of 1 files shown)


sindresorhus:

PR #290, opened 2015-06-06
Add `tabex`

puzrin:
https://github.com/nodeca/tabex

Formaly, it's not pure node project, because it works in browser. But it's very "close" to all realtime buses. I don't know why, but none of existing buses allows to share single server connection between tabs, when user opens multiple pages. And i didn't found convenient extentions for that (inspected `wormhole`, `IWC`, `crosstab` and so on). Though problem of saving server resources & client traffic is very serious.

I think project is cool, but not sure if if fits to pocily of your list. Decide yourself.


sindresorhus:
Yup, looks nice, but as you said, not a node thing.

Would be a better fit here: https://github.com/sorrycc/awesome-javascript

PR #292, opened 2015-06-11
Added marvel-api to Mad Science

DeekshithShetty:


sindresorhus:
Thanks, but not really mad science.

PR #294, opened 2015-06-12
Added geeky section to readme.md

DeekshithShetty:
Have removed marvel-api from Mad science and put in a separate section called geeky


sindresorhus:
Thanks, looks interesting, but not general enough to be included.

PR #295, opened 2015-06-12
Add ObjectEventTarget

gartz:
https://github.com/gartz/ObjectEventTarget

It worths because you can set it as prototype of constructor function and all instances will be able to listen for events you can dispatch from inside your object logic, just like you work with DOMElements, what is very well documented and reliable. And you don't need to create any special property in your object it works detecting the context of the object, and easy allow to that object be extended and still work.

Test coverage 99% and documented.


sindresorhus:
Looks useful in the browser, but in Node there's the built-in EventEmitter, so don't really see the point there. Try https://github.com/sorrycc/awesome-javascript instead.

gartz:
Thanks for recommend that other repo.

PR #296, opened 2015-06-21
Add Mitm.js.

moll:
Hey,

Mitm is an another intercepting and mocking library that I'd say has a better API that the DSL-y alternatives — one based on Node's core HTTP request and response API.

Cheers


sindresorhus:
Thanks but going to pass on this.

The readme doesn't mention how it's different (better) from existing ones like `nock` and I don't want to recommend something with an LGPL license.

moll:
Thanks for the response at least.

I thought the following did explain it well:
> Does not have any kitchen sink features or yet another API to assert on intercepted connections.

Internal DSLs have their place, but I'd limit DSLs per tests at one — an assertion library. Did you check the examples? They look like you have an plain old Node HTTP/TCP server. Mitm.js works with TCP connections, too, but that's just an extra. ;-)

The LGPL dislike is peculiar though. It merely prevents people from leeching on other people's work, but isn't as contagious as GPL is. I have an almost opposite stance — I'd not waste my time on something that someone else can take, fix or improve and _not_ share that fix.

PR #303, opened 2015-07-06
Propose adding moron.js to ORM category

r4j4h:
I encountered it in [this article](http://www.vincit.fi/blog/introducing-moron-js-a-new-orm-for-node-js/) and it is well tested and appears well put together and I feel is a worthy submission.

[Here is an issue](https://github.com/Vincit/moron.js/issues/11) regarding its similary to Bookshelf

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -455,6 +455,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 	- [Knex](http://knexjs.org) - A query builder for PostgreSQL, MySQL and SQLite3, designed to be flexible, portable, and fun to use.
 - Other
 	- [NeDB](https://github.com/louischatriot/nedb) - Embedded persistent database written in JavaScript.
+	- [Moron.js](https://github.com/Vincit/moron.js) - Declarative ORM built around [knex](https://github.com/tgriesser/knex) SQL query builder, similar to Bookshelf but with less Backbone.js style influence.
 
 
 ### Testing


koskimas:
Hi, moron.js author here. 

If this gets accepted I think a better description would be something like "An SQL friendly ORM built around [knex](http://knexjs.org) with helpers for working with documents". 

Thank you for making suggestion and pull request @r4j4h.

sindresorhus:
Not very interested in adding *yet another* ORM unless you can convince me otherwise. There's already way too many in the list: https://github.com/sindresorhus/awesome-nodejs#database

koskimas:
The [blog post](http://www.vincit.fi/blog/introducing-moron-js-a-new-orm-for-node-js/) explains pretty well why I wrote yet another ORM. One of the things the blog doesn't mention is the focus in the ES6/ES7 compatibility. Check out the [es7 example project](https://github.com/Vincit/moron.js/tree/master/examples/express-es7).

r4j4h:
@sindresorhus I highly recommend reading that blog post, it's the same one I linked to and explains where moron fits in relation to all those other ORMs. It's a response to the problem of there being too many in ambiguous states, esp. in regard to relations and dialect and opinions.

@koskimas It is clear the description is of upmost importance to identifying the core things that differentiate it from others.

Another approach that might be productive could be to update the descriptions of the existing ones in the same manner. (e.g. orm2 includes "Mongo" as if its a first class adapter but its still in beta (2 years?), does not support transactions or eager loading, and there's no issue even mentioning it with todos all over the source base). I'm not trying to call orm2 out specifically, just trying to agree with not wanting to add to a ball of mud while showing there may still be room if we make it less of a ball of mud to begin with)

sindresorhus:
Ok, looked into it some more. There are many things I like about it. Gonna pass on it for now though, as it was just open sourced and I think it could use some time to mature before recommending it here. I also think the name of it is unfortunate. Moron has some negative connotations and you're surely going to lose out on users just because of that. Happy to reconsider in a few months ;)

koskimas:
Ok, thanks for considering it. Yeah, I'm kind of surprised how many people consider the name offensive. We are probably going to change it.

gullyfoyle:
FYI, I love the name! I find it very strange that some people find it a bit 'too much'. I never realised a lot of devs were so sensitive (I'm referring to /r/node, not you @sindresorhus !). 

A funny note: in Ireland & Great Britain, the word 'mongo' is a _very_ derogatory term used to describe someone who is mentally retarded. It doesn't stop people from using it...

PR #308, opened 2015-07-09
clarified `navit` description

puzrin:


arthurvr:
I'd rather see us changing the description to what their readme provides too:

> Wrapper for PhantomJS & SlimerJS to simplify browser tests scripting.

PR #310, opened 2015-07-11
proposing math-interval-js

vluzrmos:


sindresorhus:
Thanks, but it's a bit too obscure of a use-case to be on this list. Nice work though :)

I would also consider renaming the package on npm from `math-interval-js` to `math-interval`. The `js` postfix is just moot.

PR #314, opened 2015-07-14
Add link to FlatDoc

shamsher31:
Add link to FlatDoc documentation generator


sindresorhus:
Why should I add it? I might be missing the point, but it seems like it only render the readme, which is already pretty readable on GitHub.

shamsher31:
It uses README but gives structured way for documentation by providing good navigation bar on the left,  description and details in center and code examples on right panels and everything on single page, which I think makes developer life easier.

sindresorhus:
I don't really see the point tbh, but I'm going to pass on it for the reason that it's a front-end thing and doesn't use Node in any way which is what this list is about.

PR #319, opened 2015-07-21
Added new logging library

Xalbi:
Hello, 

Why it should be included ? I think sometimes we just need a simple logger.
  
https://www.npmjs.com/package/xlblogger

Best regards.


sindresorhus:

PR #320, opened 2015-07-22
Add `mongo-url-utils`

seangarner:
[mongo-url-utils](https://github.com/seangarner/mongo-url-utils) parses url params into objects which can be passed to the mongo find method.  Cleaner (IMO) and safer than just serialising a json object into the url.


sindresorhus:
Looks useful, but a bit too obscure for this list :)

PR #321, opened 2015-07-23
added robotjs

dthree:
RobotJS is awesome, and has a lot of potential. I looked through all of the categories and couldn't really find a good place to put it, so I am proposing an "automation" category for it.


kevva:
Yup, thanks :).

PR #322, opened 2015-07-23
Add NW.js & awesome-nw (against electron)

fritx:
;) thank you!


sindresorhus:
Thanks, but I think Electron is better and this is a curation after all. Happy to consider including awesome-nw on https://github.com/sindresorhus/awesome/ though ;)

fritx:
@sindresorhus ok, let me try, thanks.

PR #324, opened 2015-07-28
Adding Native Access to Mac OS-X FSEvents

jaysingh:


sindresorhus:
Looks good, but not generally useful enough as most will want some cross-platform abstraction above it, like Chokidar or any of the other fs event libs.

PR #325, opened 2015-07-29
Adding glob

jaysingh:
Match files using the patterns the shell uses, like stars and stuff.


sindresorhus:
Already have `globby`, which is just `glob` with support for multiple patterns.

PR #327, opened 2015-07-30
Add `split2` module

LoicMahieu:


sindresorhus:
Thanks, but we already have `byline`, which does the same thing.

PR #337, opened 2015-08-22
add `url-pattern`

snd:
**[url-pattern](https://github.com/snd/url-pattern) is easier than regex string matching for urls, domains, filepaths and other strings.**

information about package heath, features and examples are at the top of the [readme](https://github.com/snd/url-pattern) followed by thorough documentation.

added it to the [text](https://github.com/sindresorhus/awesome-nodejs#text) category as  there already were some packages for dealing with urls in that category.

adding a separate *Url* category might be handy as there are already 4 (5 including this) packages dealing with urls in the list.


sindresorhus:
Can you try to shortening the description a bit?

snd:
sure. hope this is better.

PR #339, opened 2015-08-24
Update Vantage.js description due to role change.

dthree:
[Vantage](https://github.com/dthree/vantage) had all of its CLI functionality extracted into [Vorpal](https://github.com/dthree/vorpal), a framework for building immersive CLI apps.

Vantage now *extends* Vorpal and adds client / server functionality. So Vantage's tagline is now `CLI` + `SSH` + `REPL` for your live app and it fits under the heading of a realtime application.

Vorpal will fit into "Command Line Apps" later on when It's more publicized.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -87,7 +87,6 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [Standard](https://github.com/feross/standard) - JavaScript Standard Style — One style to rule them all.
 - [cpy](https://github.com/sindresorhus/cpy) - Copy files.
 - [fkill](https://github.com/sindresorhus/fkill-cli) - Fabulously kill processes. Cross-platform.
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't think that's very descriptive. Not everyone knows what SSH and REPL stands for. The previous description was IMHO better.

dthree:
Okay I see where you're coming from on that one and I agree. My main worry is that I want people to get the difference between the two libraries. Maybe I should just move it to the Realtime section and leave the title alone?

sindresorhus:
Sidenote: Both of these should probably be in the `CLI utils` section.

> My main worry is that I want people to get the difference between the two libraries.

Then come up with better descriptions. Keep in mind that the module description is the most important piece of text your write for a project. Take care to make it good.

dthree:
Haha k and got it. Closing this for now. If Vorpal gets broader adoption I'll nail down the difference later on.

PR #340, opened 2015-08-25
Add a new ORM called Toshihiko

XadillaX:


sindresorhus:
Thanks for the suggestion, but going to pass on this. There's already way too many ORMs in the list.

PR #341, opened 2015-08-25
add prettylog

dariubs:


sindresorhus:

PR #354, opened 2015-09-12
Kefir.js

santacruz123:
Kefir is a Reactive Programming library for JavaScript inspired by Bacon.js and RxJS, with focus on high performance and low memory usage.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -129,6 +129,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [Bacon.js](http://baconjs.github.io) - Functional reactive programming.
 - [RxJS](http://reactivex.io/) - Functional reactive library for transforming, composing, and querying various kinds of data.
 - [Lazy.js](https://github.com/dtao/lazy.js/) - Utility library similar to lodash/Underscore but with lazy evaluation, which can translate to superior performance in many cases.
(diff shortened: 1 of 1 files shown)


sindresorhus:
> Kefir — is a Reactive Programming library for JavaScript inspired by Bacon.js and RxJS, with focus on high performance and low memory usage.

Not really sold on that. I see no reasoning how it actually differs from the rest. No proof of higher performance or lower memory usage.

santacruz123:
It's another alternative, let others to decide whether to use it or not

santacruz123:
Comparison - http://jsfiddle.net/jL1nm3c3/5/

gartz:
I like it, very simple and well documented, I will give a try soon for a frontend project that already use some event stream based libs like https://github.com/gartz/pushStateTree (focus on handling browser navigation).

ericsaboia:
+1
it seems nice IMHO.

sindresorhus:
> It's another alternative, let others to decide whether to use it or not

That's not how it works. Just saying it's faster doesn't make it true. I can create a framework and call it "the best framework in the universe and all parallel dimensions", but that doesn't make it so. There needs to be some kind of evidence, preferably in your marketing. If I added every alternative it would no longer be a curation, and the list would lose its value.

-

I do like the API though, so happy to reconsider when you've gotten the 3.0 release out. Hopefully you'll have some better marketing and comparison by then ;)

santacruz123:
It's not my framework, dinosaur ;)

PR #356, opened 2015-09-13
Add `event-wire`

puzrin:
https://github.com/nodeca/event-wire

Decide yourself. That's a combination of mediator and responsibility chain paradygms + sugar for convenience. Simplifies creating apps with extendable logic. See [example](https://github.com/nodeca/nodeca.forum/blob/master/internal/forum/post_list.js) with before/after/on stuff.

This approach is used to build miracle nodeca app :).


sindresorhus:
Thanks for the suggestion but going to pass on this.

PR #360, opened 2015-09-23
Add subdownloader

beatfreaker:
Hi,

I have made a CLI utility to download subtitles for movies and TV series.
you just have to execute a simple command to download subtitles for one or more file at a time.
I am using it daily and its very helpful hope it can help others too

https://github.com/beatfreaker/subdownloader

Thanks :smiley:


sindresorhus:
Nice!

One improvement I would consider is to split it into a CLI module and API module. That would make the code cleaner and allow other people to use the code programatically.

Example:
https://github.com/sindresorhus/hasha-cli
https://github.com/sindresorhus/hasha

beatfreaker:
Thanks.

Will implement the suggestion give in couple of days.

PR #369, opened 2015-10-08
add node framework

azat-co:
add NodeFrameworks.com](http://nodeframeworks.com) - Curated list of hand-picked Node.js frameworks and libraries


sindresorhus:
Thanks, but I'm going to pass on this. As it stands it's just a list with random links without almost no context. It could have included descriptions for each framework and when and why one might choose to use it, and more. I also get the feeling it's only made to promote your book. I'm sure it's not, but that's the feeling I get when I see a quickly assembled site with a big red ad banner at the top.

PR #373, opened 2015-10-09
Add Vorpal

dthree:
I was going to propose this later on as it matured, but it's currently trending in the top 5 on HN, so thought now would be a good time, if that's okay with you. :)


dthree:
Thanks :D

PR #375, opened 2015-10-14
Add Sane filesystem watcher

andela-earinde:
Link to repo: [sane](https://github.com/amasad/sane)

Very simple to use.


sindresorhus:
Thanks, but `chokidar` is IMHO better and already on the list, so gonna pass on this.

PR #380, opened 2015-10-18
sails-js-in-action

san4osq:
sails-js-in-action


sindresorhus:
Thanks for the suggestion, but the book is not done, so gonna pass on this until it is.

PR #390, opened 2015-11-03
add hanging-indent

codekirei:
Small, simple module that formats a long string into a hanging-indented paragraph with newlines and indentations. Breaks lines in between words. Configurable max line length and indentation size.

I believe this functionality is unique among the modules currently in the text section.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -368,6 +368,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 - [i18n-node](https://github.com/mashpie/i18n-node) - Simple translation module with dynamic JSON storage.
 - [babelfish](https://github.com/nodeca/babelfish/) - i18n with very easy syntax for plurals.
 - [parse-columns](https://github.com/sindresorhus/parse-columns) - Parse text columns, like the output of Unix commands.
+- [hanging-indent](https://github.com/codekirei/hanging-indent) - Format a string into a hanging-indented paragraph.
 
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
What would anyone use this for? Keep in mind that this list is for Node.js, not modules in general. I can see text wrapping being useful in the terminal, but no one would use a hanging indent there.

codekirei:
I suppose it is rather niche :innocent:

The application I have in mind is printing a large number of long, unformatted strings consecutively in a CLI app. For example, a web crawler with some rudimentary parsing logic to render a page's HTML to an array of strings. Depending on what you crawl, you could end up with a very large array of strings of vastly differing lengths. Say you want to write the output of this app to a logfile or simply `console.log` to term.

If your output has consecutive very long strings, unindented wrapping would lead to a huge block of text. Depending on the length of the final wrapped line of each string in the block, it may be difficult to tell where each string in the block begins and ends. Adding an additional `\n` or some other indicator between each string solves readability, but if you have a large number of single-line strings to print as well, you just doubled the length of your output.

I thought a hanging indent would be happy medium for readability--it clearly delineates the start and end of long strings without adding a ton of blank newlines.

sindresorhus:
I still think a hanging indent looks bad. Much prefer A. But I guess this is the tab/spaces of the writer world. Gonna merge since you feel so strongly about it and the code looks pretty good.

codekirei:
:+1:

PR #397, opened 2015-11-18
Add urlshortener

G07cha:
Library for interaction with popular url-shorteners.


sindresorhus:

G07cha:
@sindresorhus, can you explain why it's isn't good enough so we can improve `urlshortener`?

PR #402, opened 2015-11-21
Add log4js

glabcn:
Add log4js. A conversion of the log4js framework to work with node.


sindresorhus:

manuc66:
Could it be reconsidered nowadays?

PR #403, opened 2015-11-22
Add Danf in web frameworks section

Gnucki:
[Danf](https://github.com/gnodi/danf) is a full-stack isomorphic OOP framework allowing to code the same way on both client and server sides. It helps you to make deep architectures and handle asynchronous flows in order to help produce scalable, maintainable, testable and performant applications.

This framework has its own paradigm and can be interesting to know.


sindresorhus:
Thanks for the suggestion, but I'm going to have to pass on this. It's hard to judge a framework without any visible usage or community. Happy to reconsider some time in the future when you've gather some traction and community around the project ;)

Gnucki:
Ok no problem! Hope I will succeed in gathering this community!

PR #406, opened 2015-11-26
Add TVSort

torrottum:
Sorts TV show files into subdirectories. Currently uses this format `Show title/Season X/Filename`.


sindresorhus:
The code itself looks pretty good, but the utility of the CLI isn't generally useful enough to be in this list (meaning the target group is too small).

I would recommend you instead submit to: https://github.com/aharris88/awesome-cli-apps ;)

torrottum:
Thanks, that actually means a lot coming from you :grinning:!
Thanks for the recommendation

PR #407, opened 2015-12-01
Add transceiver

risq:
`transceiver` is a channel based event bus, written in ES6, using promises. It implements request/reply pattern (inspired by backbone.radio, with promises). It is designed to fit with bluebird API or any promise engine (but also works without any one as a traditional event bus).

The API is fully documented and 100% test covered :)


sindresorhus:
The code itself looks pretty good, but I think event-buses is an anti-pattern, so I'm going to pass on this.

PR #411, opened 2015-12-16
Added CodeceptJS testing framework

DavertMik:


sindresorhus:
Looks interesting, but I'd like to see it mature more before adding it to the list :)

PR #413, opened 2015-12-20
Update readme.md

HendrixString:
a lightweight HTTP library for Node.js inspired by Square's Android OkHttp Builder patterns and flavours for constructing a legit HTTP requests of all types. Written entirely in EcmaScript 6 (Harmony) and transpiled with Babel.

will help any developer constructing any type of common http request intuitively. for example multipart requests.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -144,6 +144,7 @@ Inspired by the [awesome](https://github.com/sindresorhus/awesome) list thing.
 
 ### HTTP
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
:palm_tree: Thanks for the suggestion, but I'm going to pass on this.

HendrixString:
what? why? I believe this has potential educational and productivity values.
It is used in most of our products and simplifies our lives http-wise.
Is is also a humble homage to the great OkHttp of Square for Android.

SamVerschueren:
No offence to the repository, but it's not because you think it's awesome, other people think it's awesome as well. Just do another PR if the userbase/community grows.

PR #415, opened 2015-12-22
Added OS.js

junland:
Added OS.js to `Miscellaneous`


sindresorhus:
It looks cool, but while the server does indeed run on Node.js, it's just a tiny part of the project and the main part being the front-end. I'd recommend you rather submit it to https://github.com/sorrycc/awesome-javascript ;)

junland:
@sindresorhus Thanks! :)

PR #424, opened 2016-01-03
Add documentation.js

kt3k:


sindresorhus:
I don't think ESDoc is stable enough to include.

PR #427, opened 2016-01-10
adding node-ipc and node-http-server

RIAEvangelist:
recommended by a @Rokt33r on RIAEvangelist/node-ipc#56


sindresorhus:
:palm_tree: Thanks for the suggestions, but I'm going to pass on these. :)

PR #429, opened 2016-01-13
Add diff2html-cli to Command-line apps

rtfpessoa:
[diff2html-cli](https://github.com/rtfpessoa/diff2html-cli) is an easy to use command-line application that creates pretty HTML diffs from git diff outputs, allowing a better code comparison in the browser.


sindresorhus:
:+1: This would be a lot more useful if wrapped in Electron and opened in a floating window that closed when the process closed. That way I could just diff from the CLI and close with Ctrl+C in the terminal. I wouldn't use this as is as it would be too annoying to have to manually close the tab each time I wanted to diff.

sindresorhus:
Might want to also submit to https://github.com/aharris88/awesome-cli-apps

rtfpessoa:
Ok. I will look into it when I have some time.

sindresorhus:
https://github.com/dominictarr/electro would make it easy ;)

rtfpessoa:
Thanks for the help :+1:

PR #430, opened 2016-01-13
Add diff2html to Parsing

rtfpessoa:
[diff2html](https://github.com/rtfpessoa/diff2html) is an simple parsing library for git diffs that creates a JSON representation of the diff and can also generate pretty HTML.


sindresorhus:
Looks really good, but I don't think it's generic enough for this list.

Maybe try https://github.com/sorrycc/awesome-javascript.

rtfpessoa:
Ok, fair enough. I just thought since there was no other good diff parser and generator this would be a good addition.

PR #434, opened 2016-01-17
Add core-util-is to Polyfills

romankl:
`Core-util-is` is a module that extracts the v0.12 `util.is*` ([on npm](https://www.npmjs.com/package/core-util-is))


sindresorhus:
All those methods are deprecated in Node.js since they're better done in other modules, like e.g. Lodash.

PR #436, opened 2016-01-19
Added dns-zonefile

elgs:
https://github.com/elgs/dns-zonefile


sindresorhus:
When would I use something like this?

elgs:
@sindresorhus `dns-zonefile` is to parse a zone file to JSON format, and to generate a JSON file to zone file format. This is useful for parsing, generating or manipulating DNS zone files. `dns-zonefile` is used by Google Cloud Platform for their Node.js client: https://github.com/googlecloudplatform/gcloud-node.

sindresorhus:
I read the readme on what it does, I'm just unsure in what situation I would use this. I'm trying to keep the list to what people would frequently need and I can't see what I would use this for in my projects.

sindresorhus:
Gonna pass on it for the above reason. Sorry. I need to be picky on what I include otherwise it's no better than the npm search.

PR #437, opened 2016-01-19
Add JavaScript Air

kentcdodds:


sindresorhus:
How exactly is this specifically related to Node.js? From looking at the website it seems more JS generic and should probably be added to https://github.com/sorrycc/awesome-javascript instead.

kentcdodds:
Perfect. Thanks for directing me to something more appropriate for it. :-)

PR #438, opened 2016-01-19
Add cross-env

kentcdodds:


sindresorhus:
The description needs work.

kentcdodds:
Yeah, I struggled with it. How about:

```
Set environment variables in scripts with unix style and it still works on Windows.
```

I'm not a huge fan of that either... :-/

sindresorhus:
> Set environment variables cross-platform.

sindresorhus:
You should also add it here: https://github.com/bcoe/awesome-cross-platform-nodejs

SamVerschueren:
```
Expose unix style environment variables cross-platform
```

Or what Sindre said :)

kentcdodds:
Updated. Thanks!

PR #439, opened 2016-01-22
Added json-configurator

cmartin81:
A nice tool for those how needs to have have different variables in different environment. (test variabels vs prod variables)


sindresorhus:

PR #440, opened 2016-01-25
add multiline-billboard

codekirei:
[multiline-billboard](https://github.com/codekirei/multiline-billboard) is a simple little module to turn a string or array of strings into a nicely-styled billboard.

Potentially useful for CLI apps that log to stdout to:

- call attention to something
- embellish a header

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -219,6 +219,7 @@
 - [multispinner](https://github.com/codekirei/node-multispinner) - Multiple, simultaneous, individually controllable CLI spinners.
 - [omelette](https://github.com/f/omelette) - Shell autocompletion helper.
 - [cross-env](https://github.com/kentcdodds/cross-env) - Set environment variables cross-platform.
+- [multiline-billboard](https://github.com/codekirei/multiline-billboard) - Turn text into a fancy billboard for the terminal.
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Seems exactly the same as https://github.com/sindresorhus/boxen, which I was planning to do add to the list soon.

codekirei:
Oh cool! Haven't seen boxen before. Definitely overlapping use cases. From a brief poke around the docs and source, I do see a few unique implementation differences:

**boxen has:**

- configurable margin
- configurable padding

**multiline-billboard has:**

- configurable border characters
- configurable text justification (left/right/center)

Neither of these feature sets would be difficult to implement in the other module. I do agree that only one of the modules should be included in the list, though &mdash; so if you want to close this in favor of your module I totally understand :smile:

LMK if you're interested in having configurable border chars or text justification in boxen &mdash; would be happy to make a PR!

sindresorhus:
Sorry, closing in favor of `boxen`...

> LMK if you're interested in having configurable border chars or text justification in boxen — would be happy to make a PR!

The former was added by a contributor, would be happy to accept a PR for the latter :)

PR #442, opened 2016-01-26
add rfpify

SamVerschueren:
Saw `pify` in the list, so added [rfpify](https://github.com/samverschueren/rfpify) as well. Don't feel bad if you reject the PR :).


SamVerschueren:
Thanks for the merges! :tropical_drink:

PR #455, opened 2016-02-09
Add huzzah logger

btd:


sindresorhus:
Thanks for the suggestion, but I'm going to have to pass on this. It's hard for me to judge a logger without any community. Happy to reconsider some time in the future when you've gathered some traction and community around the project ;)

btd:
Fair enough.

PR #460, opened 2016-02-15
Add gradle-to-js

karllindmark:
Hello @sindresorhus! Would you care to include this gradle build file parser library under the "parsing" section? The first (and probably the last) of its kind, heh.


sindresorhus:
Thanks for the suggestion, but this is unfortunately too obscure to include in the list.

PR #462, opened 2016-02-23
Add Cash

dthree:
[Cash](https://github.com/dthree/cash) implements popular Unix / Linux commands such as `ls`, `grep` `less` using straight ES6, installed through Node.


dthree:
Thanks! You're awesome. :+1:

PR #464, opened 2016-02-27
Add ygor.

shannonmoeller:
A promising task runner. Fills a gap between `npm run` and tools like `gulp` and `grunt`. Promotes `async` and `await` for flow control.

https://github.com/shannonmoeller/ygor

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -252,6 +252,7 @@
 - [strong-build](https://github.com/strongloop/strong-build) - Build a node app package and prepare to deploy it as a package to production or use git to commit to a deploy branch.
 - [grunt](http://gruntjs.com) - Task runner that can perform repetitive tasks like minification, compilation, unit testing, linting, etc.
 - [start](https://github.com/start-runner/start) - Simple tasks runner powered by composable functions and promise chaining.
+- [ygor](https://github.com/shannonmoeller/ygor) - A promising task runner for when `npm run` isn't enough and everything else is too much.
 
 
 ### Hardware


sindresorhus:
While I think think gulp/grunt does too much, this is doing IMHO too little.

> Need to run asynchronous tasks? Cool. Use the async and await keywords for flow control.

This is also deceiving as it doesn't mention you'd have to run it through something like Babel first.

-

I'm gonna pass on it for now. Work on it some more and I'd be happy to reconsider in the future ;)

shannonmoeller:
> this is doing IMHO too little

Your list, your opinion; totally fair.

The simplicity is intentional though. I feel like Gulp, et al, have lost sight of their [original goal](https://github.com/gulpjs/gulp/issues/193#issuecomment-33299786). Proponents of npm scripts go in the opposite extreme by assuming that the cli apis are sufficient. Ygor sits in the middle by assuming you want to use JS apis. I have used the equivalent of this script in a number of client projects, got tired of the copy-pasta, and published it.

> it doesn't mention you'd have to run it through something like Babel first

I could definitely expand the README and I need to publish the latest copy which includes an example of using `babel-node` in addition to [esprev](https://www.npmjs.com/package/esprev). Jumped the gun a bit in assuming that async/await are the future and no longer needed a fuller explanation.

> Work on it some more

I've been toying with bundling globby, fs-promise, and fs-extra. Is that the sort of work you're looking for?

sindresorhus:
> I've been toying with bundling globby, fs-promise, and fs-extra. Is that the sort of work you're looking for?

No. Definitely not. I think it's in the correct direction. Just felt a bit incomplete/immature. But you know what. I'm gonna merge.

shannonmoeller:
Oh, wow. Thanks!

shannonmoeller:
Especially for reading to the end of the long reply when you had already closed the issue. You rock.

PR #465, opened 2016-02-27
Add cli-columns.

shannonmoeller:
Loggable columnated text lists. Like `ls` output. Unicode and ANSI safe.

screenshot

https://github.com/shannonmoeller/cli-columns

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -236,6 +236,7 @@
 - [progress](https://github.com/tj/node-progress) - Flexible ascii progress bar.
 - [cli-cursor](https://github.com/sindresorhus/cli-cursor) - Toggle the CLI cursor.
 - [columnify](https://github.com/timoxley/columnify) - Create text-based columns suitable for console output. Supports cell wrapping.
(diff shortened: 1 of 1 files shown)


sindresorhus:
How is this any different from `columnify`?

shannonmoeller:
Despite its name, `columnify` formats tabular data and has much more in common with `cli-table`. Given a list of strings, `columnify` just dumps them out in a single column (including the array's `length` property):

```js
import columnify from 'columnify';
console.log(columnify(['g', 'a', 'b', 'c', 'd', 'e', 'f']));
```

```
0
a
b
c
d
e
f
g
h
```

`cli-columns` is not for tabular data. It takes a list of strings and displays them in sorted balanced columns depending on the width of the console window exactly like the [Unix `column`](http://linux.about.com/library/cmd/blcmdl1_column.htm) command.

```js
import columns from 'cli-columns';
console.log(columns(['g', 'a', 'b', 'c', 'd', 'e', 'f']));
// assume terminal window is 12 or so columns wide
```

```
a  d  g
b  e  h
c  f
```

dthree:
Might use this in Vorpal.

shannonmoeller:
Thank you!

PR #466, opened 2016-02-27
Add require-glob.

shannonmoeller:
Requires multiple modules using glob patterns and combines them into a nested object.

https://github.com/shannonmoeller/require-glob


sindresorhus:
No thanks. I don't think glob requiring is a good pattern.

shannonmoeller:
Thank you for your consideration.

PR #467, opened 2016-03-02
nodejieba

yanyiwu:
The Jieba Chinese Word Segmentation Implemented By Node.js .


sindresorhus:
Looks interesting, but it's a bit too obscure of a use-case for this list, which tries to common things ;)

yanyiwu:
Because it is for chinese word segmentation.
English [README](https://github.com/yanyiwu/nodejieba/blob/master/README_EN.md)

PR #473, opened 2016-03-13
Add Hexu (https://www.npmjs.com/package/hexu) [request]

KingPixil:


sindresorhus:

Its use-case is just too narrow.

KingPixil:
Got it, its fine :)

PR #476, opened 2016-03-17
Add macaca

xudafeng:
Macaca is a solution for automation test which based on Node.js.


sindresorhus:
Thanks for the suggestion, but I don't think this is ready for public consumption. Visiting the repo, I have no idea what it actually does, looking at the [docs](http://macacajs.github.io/macaca/guide.html) and they're "under construction".

PR #477, opened 2016-03-20
Add trymodule

VictorBjelkholm:
Made a simple CLI for trying out different node modules directly from the terminal. Helps developers to evaluate packages without having to add to an already existing project so hopefully this is helpful for many people.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -155,7 +155,7 @@
 - [vaca](https://github.com/sindresorhus/vaca) - Get a random ASCII 🐮.
 - [gh-home](https://github.com/sindresorhus/gh-home) - Open the GitHub page of the repo in the current directory.
 - [npm-home](https://github.com/sindresorhus/npm-home) - Open the npm page of a package.
-
+- [trymodule](https://github.com/VictorBjelkholm/trymodule) - Tool for trying out different npm modules in the terminal.
 
 ### Functional programming
 


sindresorhus:
> https://github.com/VictorBjelkholm/trymodule/blob/2030806d790213d426456959fdc7a69773a4c6f7/index.js#L9

Use `os.homedir()`.

> https://github.com/VictorBjelkholm/trymodule/blob/2030806d790213d426456959fdc7a69773a4c6f7/index.js#L59

Oh noes. Please don't. Never do string concat for commands, use `execFile` instead. Never pass environment variables directly to `rm` without sanitation, better yet, never spawn `rm`. Use a JS tools, like `rimraf` or even better [`del`](https://github.com/sindresorhus/del).

> https://github.com/VictorBjelkholm/trymodule/blob/2030806d790213d426456959fdc7a69773a4c6f7/index.js#L5

You might want to reconsider using `colors`. It's a bad unmaintained module from user hostile author. Check out `cli-colors` or `chalk` instead.

VictorBjelkholm:
@sindresorhus thanks for merging but even more thanks for the feedback! It's awesome!

I'll look into solving these issues. Thanks again for taking the time!

sbruchmann:
@sindresorhus wrote:
> You might want to reconsider using colors. It's a bad unmaintained module from a user-hostile author.

What’s the problem with Marak?

PR #482, opened 2016-03-26
Add OracleDB to Database Drivers

AnatoliyGatt:
oracledb - Oracle Database driver by Oracle Corp.


sindresorhus:
Thanks for the suggestion, but I don't think OracleDB is common enough to be in the list. Users that need it can easily find it on npm.

AnatoliyGatt:
@sindresorhus Fair enough :+1:

AdeelK93:
Oracle is one of the most popular (if not the most popular) enterprise databases out there though, and it's Node support is pretty solid. It also integrates well with Knex, another library you have on your list. I'd consider adding it to the list, oracledb does make Node more awesome

PR #487, opened 2016-04-07
Add mssql to database drivers

patriksimek:
mssql is a well docummented production-ready SQL Server client.


sindresorhus:
Thanks for the suggestion, the library and docs looks excellent, nice work, but I don't think Microsoft SQL Server is common enough to be in the list. Users that need this can easily find it on npm ;)

PR #490, opened 2016-04-16
KAJS - Dead-simple Javascript

KingPixil:
I thought my KAJS javascript engine would be a good fit for "Templating", although not 100% sure...

KAJS has many features, such as implicit variable declaration, an ruby-like @ shortcut for `this`, and optional `()` for functions!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -704,6 +704,7 @@
 
 ### Miscellaneous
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the suggestion, but I don't think this is a good fit :)

KingPixil:
It's ok :)

Just keep developing...

PR #505, opened 2016-05-10
Add Objection.js

b3njamin:
Is a well documented and test orm https://github.com/Vincit/objection.js


sindresorhus:
How is it better/different than the long list of other ORM in this list?

PR #506, opened 2016-05-11
Add js-validator-livr

pterolex:
Add LIVR validator –  Lightweight validator supporting Language Independent Validation Rules Specification (LIVR)
https://github.com/koorchik/js-validator-livr


sindresorhus:
Thanks for the suggestion, but there are already many validations modules in the list, all pretty much doing the same thing. I don't see what this brings to the table for Node.js. With Node.js you're running JS on the back-end and front-end, so no point in having language independent rules. Also doesn't look very actively maintained.

PR #508, opened 2016-05-12
Added no-more-secrets app

ptkach:
Updated Readme with new command-line app


sindresorhus:
Fun gimmick 😝, but the Command-line section is for actually useful things that solve real problems.

I'd try submitting it to https://github.com/aharris88/awesome-cli-apps instead ;)

PR #509, opened 2016-05-13
add elasticdump

manbusky:
A moving and saving indicies tool for elasticsearch.


sindresorhus:
That is very specific to Elasticsearch. This list focuses on more general tools that would benefit most users. I recommend submitting to https://github.com/dzharii/awesome-elasticsearch instead ;)

PR #511, opened 2016-05-16
Add Mostly Node podcast

NikitaLiashenko:
Mostly Node podcast was added

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -812,6 +812,7 @@
 ### Podcasts
 
 - [NodeUp](http://nodeup.com)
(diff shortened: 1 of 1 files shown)


kevva:
From the contribution guidelines:

> The pull request should have a useful title and include a link to the package and why it should be included.

`Updated readme.md` is not useful.

NikitaLiashenko:
It should be included because this is one more podcast about Node and now new NodeUp episodes rarely come after the 100 episode.

PR #513, opened 2016-05-18
Add "Secure Your Node.js Web Application" book

NikitaLiashenko:
I have read this book recently and I think it's the best thing, where you can find security tips and tricks with good explanation.


sindresorhus:
Thanks :)

PR #516, opened 2016-05-22
Add Lux

zacharygolba:
[Lux](https://github.com/postlight/lux) is an opinionated web framework that has it's roots in frustration over how much we tend to bike-shed over "solved problems" when building a large API with Node.js.


sindresorhus:
Thanks for the suggestion, but I'm going to have to pass on this. It's hard for me to judge a framework without any community. The project isn't even 0.0.1 yet. This list is for proven solutions. Happy to reconsider some time in the future when you've gathered some traction and community around the project ;)

zacharygolba:
Understood. Thanks for taking the time to explain. :)

PR #517, opened 2016-05-23
Add clui module

Arkotek:


sindresorhus:
Thanks, but it doesn't seem very actively maintained and I think it tries to do too much. Those widgets are better as separate modules, which are already in the list.

PR #518, opened 2016-05-31
Add mississippi - collection of useful stream modules

darrrk:


sindresorhus:
Please read and adhere to the [contributing guidelines](../blob/master/contributing.md).

darrrk:
Updated to be on bottom

kevva:
Not sure why you would need a module like that will all those dependencies (most which are already included on this list),

sindresorhus:
Yeah, it's a very useful guide on which modules to use, but I don't want to recommend users using just an exported collection of existing modules.

PR #520, opened 2016-06-04
Add kad

omphalos:


sindresorhus:
How is it mad science?

omphalos:
Well I think running a DHT over WebRTC qualifies as "mad science" as much as webcat, bitcoinjs, or webtorrent - decentralized data stores of different sorts running in the browser (although it's lower level than bitcoin or torrent for sure).

omphalos:
Thank you! 👍

PR #524, opened 2016-06-07
Add Kiss CLI

sixertoy:
Kiss! :
- is IDEs agnostic
- share snippets in a project
- create multiple files based on single/multiple snippets in one line
- kiss, the name, is keyboard compliant

```bash
kiss src/views/view.jsx src/index.ng
```
> will output a simple React's `src/views/view.jsx` and an Angular's module in `src/index.ng`, using fs-extra to create folders hierarchy

https://github.com/sixertoy/kiss-cli/


sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. The CLI tool list is growing too large and I need to be very picky about what I add there.

sixertoy:
Thank you for taking time :)

PR #526, opened 2016-06-08
Add `simple-get`

CodeDotJS:


sindresorhus:
There's already many libs like this in the list. This one doesn't seem to bring anything new to the table.

CodeDotJS:
Hi @sindresorhus 
Some tools and modules looks awesome but I don't know whether they deserve to be in the list or not. 

But thanks for the guidence. :)

PR #530, opened 2016-06-14
Add 'Awesome Node.js @LibHunt' to the Discovery section

StanBright:
It is based on the awesme-nodjes list here plus some helpful additions. For example: ordering by popularity, dev activity, trends and the ability to compare the different libs.


sindresorhus:
No thanks :)

StanBright:
Fair enough, you are the person taking the decision :). Yet, it would be very polite of you to shed some light why it isn't suitable?

Thanks!

PR #531, opened 2016-06-14
Add RegEx-DoS

jagracey:
It is important to prevent ReDoS attacks. Here is a CLI tool to identify possible regex denial of service (ReDos) vulnerabilities in your project.


sindresorhus:
Good idea, though it would be a lot more useful as an ESLint plugin or bundled in https://github.com/nodesecurity/nsp. I'm probably not going to run a separate tool for this, but I use the former tools.

PR #535, opened 2016-06-17
Add awesome CSS linting utility.

evilebottnawi:
https://github.com/stylelint/stylelint


sindresorhus:
Stylelint is great, but I don't think it fits into this list, even though it's made in Node.js, I'm trying to keep the list focused on tools that can be useful for Node.js devs.

Try adding it to https://github.com/aharris88/awesome-cli-apps instead ;)

PR #540, opened 2016-06-28
Add Aerospike DB client

jhecking:
Aerospike is a distributed NoSQL database and key-value store. The aerospike npm package provides a Node.js client library to connect to an Aerospike server cluster. The client library and the Community Edition of the server are open source.

Aerospike Homepage: http://www.aerospike.com/
npm package: https://www.npmjs.com/package/aerospike
Github repo: https://github.com/aerospike/aerospike-client-nodejs


sindresorhus:
Never heard about Aerospike before, but looks interesting.

For anyone else looking, found some info here: http://stackoverflow.com/questions/24482337/how-is-aerospike-different-from-other-key-value-nosql-databases

PR #545, opened 2016-07-08
Added NW.js

TheJaredWilcurt:


sindresorhus:
Duplicate of https://github.com/sindresorhus/awesome-nodejs/pull/322.

TheJaredWilcurt:
>"Thanks, but I think Electron is better"

Can you explain anything about Electron, besides the name/logo and marketing, that is better than NW.js?

* Been around 5 years longer
* more features
* more stable
* updated within 24 hours of new chromium and node releases
* better documentation
* uses less resources
* smaller dist files
* supports about twice as many OS's
* way easier to build with
* fully capable of anything electron can do

Am I missing something? Why on earth would anyone prefer the knockoff over the better original?

Electron is actually made by an intern that worked on NW.js. It even shares files that are almost identical with the exception of variable name changes. It's just a cheap knockoff with a marketing team behind it.

PR #546, opened 2016-07-09
Lock your screen via CLI

michaelzoidl:
Hi, i've written some CLI tool which can lock your screen via a simple `$ lockscreen` command.
It's useful when you're working in an coworking office where you have/should lock your screen when leaving for e.g. lunch. 

I know you can use a shortcut on mac or activate the keychain thing in the tray-bar where you can find some "Lock Screen" function... but: _Command Line is Life, Command Line is Love. <3_ 

The package [michaelzoidl/lockscreen-cli](https://github.com/michaelzoidl/lockscreen-cli) is not TDD because it builds on [michaelzoidl/lockscreen](https://github.com/michaelzoidl/lockscreen) which is TDD.


sindresorhus:
Thanks for the suggestion, but I don't think this is generally useful enough. I'm trying to contain the list as it's growing too large. I would instead recommend submitting to https://github.com/aharris88/awesome-cli-apps.

PR #553, opened 2016-07-16
Adding Fruit ORM

Javascipt:


sindresorhus:
Thanks for the suggestion, but there are already too many ORMs in the list and this one is missing a community, activity, and has too high of a [bus factor](https://en.wikipedia.org/wiki/Bus_factor).

Javascipt:
Can you give me an advise on how to turn this package into a better one? How to find contributors and how to avoid the bus factor you mentioned ?

PR #555, opened 2016-07-24
Adding manakin to the list of command-line utilities

vitaly-t:
[Manakin](https://github.com/vitaly-t/manakin) - adds prime colors to your Node.js console output, in the most simple and safe way.

* It can set colors either locally or globally (by overriding `console`)
* No need to mess with colors, very easy to use
* Consistent formatting with the standard `console` output
* Consistent use of `stdout` and `stderr`
* Super-easy extensibility

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -266,6 +266,7 @@
 - [yargs](https://github.com/yargs/yargs) - Command-line parser that automatically generates an elegant user-interface.
 - [liftoff](https://github.com/js-cli/js-liftoff) - Bootstrapping tool for building command-line app.
 - [listr](https://github.com/samverschueren/listr) - Terminal task list.
+- [manakin](https://github.com/vitaly-t/manakin) - Prime colors for your console — quick & safe.
 
 
 ### Build tools


sindresorhus:
🌴 Gonna pass on this. I don't think colors are a good way to differentiate logging types. For example, in many environment where you care about the logs, colors are not supported. Overriding the global `console` object is also not something I would recommend people do. There are already text coloring libs and logging libs in the list that solves the same problem better.

vitaly-t:
>There are already text coloring libs and logging libs in the list that solves the same problem better.

This one is a unique library that makes the `console` override safe:

>This library focuses on providing consistent output format for console methods after applying the colors, in contrast to generic color libraries that only apply colors to text.

i.e. it is the only library that provides formatting 100% consistent with that of Node.js, for the complete syntax supported:

* with a list of parameters: `console.log(obj1 [, obj2, ..., objN]);`
* with message formatting: `console.log(msg [, subst1, ..., substN]);`

It is also the only library to colorize all your console output with a single line of code.

PR #557, opened 2016-07-26
Added fnbpasswordvalidator

FelipeNBrito:
Make sure youre users have a secure password, checking for uppercase/lowercase letters, numbers, special characters, prohibited words (e.g. their name or email address) and min/max password length.


sindresorhus:
You shouldn't check single metrics like these. They cause way too many false positives and end up just annoying users. Instead you should check the combined difficulty of a password and entropy. See https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/ and https://xkcd.com/936/

PR #558, opened 2016-07-27
Add rill-js webframework

DylanPiercey:


sindresorhus:
Thanks for the suggestion, but I'm going to have to pass on this. It's hard for me to judge a framework without any community. The project also has too high of a [bus factor](https://en.wikipedia.org/wiki/Bus_factor). Happy to reconsider some time in the future when you've gathered some traction and community around the project ;)

DylanPiercey:
@sindresorhus, thanks for checking it out. I'm curious what steps you think I should take to improve the framework? Perhaps it's just something that's too niche.

PR #559, opened 2016-07-28
Add npm-register

dickeyxxx:


sindresorhus:
This would be a better fit on https://github.com/sindresorhus/awesome-npm ;)

PR #560, opened 2016-08-01
Move agenda to job queues section

Atrox:
Hey,

I just noticed that agenda is under "Miscellaneous" instead of "Job Queues".
If there is a specific reasoning behind this I would love to know.

Thank you,
Atrox


sindresorhus:
> If there is a specific reasoning behind this I would love to know.

Probably because the Job Queues section didn't exist then. You could always just check with `git blame` ;)

PR #564, opened 2016-08-08
Added websiteWrapper

RobertJGabriel:


sindresorhus:
This list is for Electron related things, not NW.js.

PR #570, opened 2016-08-31
Update readme.md

TeddyHsieh:
Add a practice article to the article section. 
If you find it useful, please merge it.
Thanks!!


sindresorhus:
Maybe if it was on Medium or a personal blog, but the article feels more like a promotion for Codementor. When you go to article it loads up a fullscreen ad for Codementor. There are also some anti-patterns and false statements in that article. There are lots of other better articles on this topic. Thanks though :)

PR #572, opened 2016-09-03
Add nodejs-traceroute to Network section

zulhilmizainuddin:
Link: https://github.com/zulhilmizainuddin/nodejs-traceroute

Execute and retrieve results of tracert in Windows and traceroute in Linux from a Node.js program.


sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. It's not something I'd imagine the majority would need. Code looks good though :)

PR #577, opened 2016-09-11
Add lychee.js to README

cookiengineer:
Would be awesome to get lychee.js on this list!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -106,6 +106,7 @@
 - [limdu](https://github.com/erelsgl/limdu) - Machine-learning framework.
 - [Cytoscape.js](http://js.cytoscape.org) - Graph theory (a.k.a. network) modeling and analysis.
 - [kad](https://github.com/kadtools/kad) - Kademlia distributed hash table.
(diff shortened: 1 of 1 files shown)


sindresorhus:
I've been digging through its docs for 15 minutes now, and while it sounds very ambitious, it's not really selling me on anything. The webpage failed to answer the most important question: "What problem does it solve?". Self-building and self-learning software sounds great in theory, but from experience it's like the IoT trend, more a hype than actually a practical solution. It claims "Next-Gen Isomorphic Application Engine", but how exactly does it solve problems better than today's solutions? I see a lot of fancy words, but little docs and actual proof on what it's for and why I should care. I think it's an awesome idea, but I'm gonna pass on it for now as it's not something that is approachable and usable enough for this audience. Keep up the great work :)

cookiengineer:
> how exactly does it solve problems better than today's solutions?

`lycheejs-fertilizer auto /projects/boilerplate` - ready for 37 platforms, including SDL, OpenGL, WebGL, node, nwjs. All architectures, all operating systems.

> but little docs and actual proof 

What do you mean by that? Please explain what is missing, I'm eager to find out.

PR #578, opened 2016-09-12
add timeago.js & pys

hustcc:
[timeago.js](https://github.com/hustcc/timeago.js) is a tiny(~1.75kb) library used to format date with `*** time ago` statement. eg: '3 hours ago'. No dependences & localization & tiny.

[pys](https://github.com/hustcc/pys) is a ~700b javascript library to engance String.substring / Array.slice with python slice style.


sindresorhus:
Thanks for the suggestion, but I think https://github.com/sorrycc/awesome-javascript would be a better fit for these :)

PR #581, opened 2016-09-18
feat(web-frameworks): include Nodal

ahmadnassri:
Nodal is a web server and opinionated framework for building data manipulation-centric (Create Read Update Destroy) API services in Node.js for web, mobile or IoT apps.


sindresorhus:
I like the idea, but seems the project has stagnated since May and it has a pretty low [bus factor](https://en.m.wikipedia.org/wiki/Bus_factor). It's hard for me to judge a framework without any community and visible projects/companies using it. Happy to reconsider some time in the future when it's gathered some traction and community around the project ;)

PR #583, opened 2016-09-20
Add GitHub-Check-CLI to Command-Line Utilities

sahildua2305:
https://github.com/sahildua2305/github-check-cli


kevva:
This should be listed under command line applications rather than utilities.

sahildua2305:
Missed that section. Sorry. Fixed! @kevva

sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. The CLI tools list is growing too large and I need to be very picky about what I add there.

PR #584, opened 2016-09-21
Roadiz CMS added

faisalhmohd:


t1st3:
Roadiz does not seem to be built with Node.js, but with PHP. That's why it does not belong to this list.

faisalhmohd:
Right. Just noticed it, my bad!

PR #585, opened 2016-09-23
readme: add thecat: Simple cat server

coderaiser:
TheCat

Added [thecat](https://github.com/coderaiser/thecat): 🐱 Simple cat server. Get new cat on every request.


sindresorhus:
Thanks for the suggestion. This list is mature now and the bar for adding new things is very high. I'm unfortunately going to pass on this as I don't think it's something people would actually use. Hope you understand. Funny idea though 😝

Issue #588, opened 2016-09-28
Add promise-memoize?


puzrin:
https://github.com/nodeca/promise-memoize is it good for this list?

- There are a lot of sync memoizers, but almost none for `Promise`-returning functions with good cache strategies
- There is https://github.com/medikoo/memoizee universal solution, but
  - it's very complicated inside (because tries to do everything - sync / cb / promises) and is more big in browser
  - my practice shows that after migrating from callbacks to (co + yield) in node 4+, promise-fn memoiser is enougth for everything.

I'd suggest to add both. `memoizee` is universal, and `promise-memoize` is just clear and awesome :) .

kevva:
[`mem`](https://github.com/sindresorhus/mem) is on the list under the misc section.

puzrin:
@keva, `mem` is just a sync cache with expire:

- It's not useable for `Promise`, because in most case you should not cache errors
- Most async caches do not care about preferch and put cache to cold state on expire.

So, `mem` is for completely different things.

sindresorhus:
@puzrin For `promise-memoize`, sure, do a PR.

PR #589, opened 2016-09-29
Add `promise-memoize`

puzrin:
#588 

I've added link under `mem` now. May be it could be better in promise section https://github.com/sindresorhus/awesome-nodejs#control-flow, let me know if i should move.


sindresorhus:
I think it would be better placed in the promise section.

puzrin:
Updated.

PR #590, opened 2016-09-30
adding good-env

recursivefunk:
`good-env` is a smart environment variable parser


sindresorhus:
Thanks for the suggestion. This list is mature now and the bar for adding new things is very high. I'm unfortunately going to pass on this as I don't think it's something the majority would need. Hope you understand :)

PR #596, opened 2016-10-07
New template engine suggestion

tiborsaas:


sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. Porting something from PHP is not a great starting point and it doesn't look very well maintained. I would instead recommend https://mozilla.github.io/nunjucks/, which is similar.

tiborsaas:
LOL

PR #601, opened 2016-10-24
add detect-port

xudafeng:
https://www.npmjs.com/package/detect-port

detect port with ease, dependents number 102


xudafeng:
@sindresorhus

sindresorhus:

PR #602, opened 2016-10-24
Added 'switchit' - cli app framework

israelroldan:
switchit is a framework to build modern OOP-based, promise-aware command-line applications and I believe it should be included in the awesome-node list! :)


sindresorhus:
Nice work, but too much OOP for my taste. Seems like classes and inheritance are used in many places just for the sake of it. I personally much prefer [`yargs`](https://github.com/yargs/yargs/).

PR #605, opened 2016-10-31
Add lowdb

typicode:
Hi,

[Lowdb](https://github.com/typicode/lowdb) is a small database built on Lodash API. It's simple and works well when you need a local database for a CLI, small server or Electron app.

It's also flexible, you can use custom format, storage, etc... and add methods.


typicode:
Thank you :)

PR #607, opened 2016-10-31
Add node-uuid

john-pettigrew:
Added a great library for generating UUIDs.


sindresorhus:
Looks like one of its forks are better maintained: https://github.com/defunctzombie/node-uuid

I don't see the immediate gain of having this in the list. It's easily searchable: https://npms.io/search?q=uuid

PR #608, opened 2016-10-31
Add superagent

john-pettigrew:
Added a library for really easy HTTP test assertions.


sindresorhus:
Thanks, but I'm gonna pass on this. I've used SuperTest before and it's API is just too magic. `.expect()` is way too overloaded and it makes it really hard to read the assertions. There are better HTTP assertion libs on npm.

ewnd9:
@sindresorhus A part of your reasoning for declining #608:

> I've used SuperTest before and it's API is just too magic. .expect() is way too overloaded and it makes it really hard to read the assertions

But `supertest` is still valuable just for wrapping an http server and sending requests.

E.g. there is the ["Endpoint testing"](https://github.com/avajs/ava/blob/master/docs/recipes/endpoint-testing.md) recipe in `ava` docs which includes `supertest` but uses `ava` assertions.

Maybe it might still be added with a note about avoiding the builtin assertions?

PR #609, opened 2016-11-01
Add make-license

accraze:
Added a CLI that generates LICENSE files for your open source projects.


sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. The CLI tools list is growing too large and I need to be very picky about what I add there.

PR #615, opened 2016-11-04
Add krakenjs

flybayer:


sindresorhus:

PR #616, opened 2016-11-09
Add IMObject class

ReedD:
IMObject is a class that allows you to cast standard JavaScript objects into deep/nested immutable objects. This class is useful for instantiating configuration objects or similar such constants that should never be altered. Once set, properties cannot be added or altered without an exception being thrown.


sindresorhus:

PR #619, opened 2016-11-11
Add pglistend in the database section

kabirbaidhya:
A lightweight Postgres LISTEN Daemon built on top of [node](https://nodejs.org/en/), [node-postgres](https://github.com/brianc/node-postgres) and [systemd](https://wiki.debian.org/systemd).

It's a very simple yet generic daemon application that could be used in any project that makes use of Postgres' `LISTEN`/`NOTIFY` feature.

It runs as a background process that does `LISTEN` on the configured channels on a database and allows to perform custom actions on receiving [`NOTIFY`](https://www.postgresql.org/docs/9.1/static/sql-notify.html) signals on those channels.

Find more information here in the [README](https://github.com/kabirbaidhya/pglistend).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -590,6 +590,7 @@
 - Other
 	- [NeDB](https://github.com/louischatriot/nedb) - Embedded persistent database written in JavaScript.
 	- [Lowdb](https://github.com/typicode/lowdb) - Small JavaScript database powered by Lodash.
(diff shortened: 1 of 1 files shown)


sindresorhus:
Nice work! Bonus points for creating a good tutorial. It's unfortunately too specific to be on this list. We can't include everything related to databases, just the most common things, which are usually adapters and ORMs. The rest is searchable on npm.

I would encourage you to submit it to https://github.com/dhamaniasad/awesome-postgres :)

kabirbaidhya:
Makes sense. Thanks anyway :)

PR #620, opened 2016-11-15
Add git-guppy commandline app

therealklanni:
[git-guppy](https://github.com/therealklanni/git-guppy) is a tool that works in conjunction with your gulp setup to automatically configure git hooks that trigger gulp tasks. 

> Simple git-hook integration for your gulp workflows.

It's a mature/stable project that's being used by a lot of projects in production-ready situations.

It's not quite a commandline util, not quite a build tool, and it has standalone terminal commands (optional install). So not easy to categorize, but it seems that commandline app is the closest fit.


sindresorhus:
Looks useful, but I prefer adding only plain Node.js modules here, not tied to a specific ecosystem like Gulp.

I would suggest writing a recipe on how to use it for https://github.com/gulpjs/gulp/tree/master/docs/recipes It would give it a lot of eyeballs. I'm talking from experience with doing that.

therealklanni:
@sindresorhus thanks for the tip!

PR #626, opened 2016-11-29
Add Saker

eshengsky:
Add Saker template, thanks!


sindresorhus:

PR #587, opened 2016-09-28
Add oboe into HTTP section

ecancino:
Link: http://oboejs.com/

Oboe.js is an open source Javascript library for loading JSON using streaming, combining the convenience of DOM with the speed and fluidity of SAX.

It can parse any JSON as a stream, is small enough to be a micro-library, doesn’t have dependencies, and doesn’t care which other libraries you need it to speak to.

We can load trees larger than the available memory. Or we can instantiate a classical OOP models from JSON, or completely transform your JSON while it is being read.

PR #639, opened 2016-12-19
Add oboe into streams section

JuanCaicedo:
Basically reopens #587 now that I'm actively working on oboe.js. I've put oboe into the streams section as opposed to HTTP. I did that because while oboe can make HTTP requests, in node it is more abstract and able to handle any readable stream. In fact, I think a preferred interface is to make a request and pass the response stream to oboe. It's worth noting that on the browser side, right now oboe just has the ability to make the requests on its own, so it would fit into the HTTP section


JuanCaicedo:
@sindresorhus is there something else you'd like to see from this PR or from oboe before you feel ok merging?

sindresorhus:
Good intent, but you just started. My argument still stands. We can revisit in 6 months and see if it's still actively maintained then.

PR #643, opened 2016-12-28
Add python-range package

Gothdo:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


intrueder:
Looks overcomplicated, way more complicated than `_.range` https://github.com/jashkenas/underscore/blob/master/underscore.js#L717
Did not test though

Gothdo:
@intrueder My range has lazy evaluation, see https://github.com/Gothdo/range#lazy-evaluation

sindresorhus:
I agree. Looks over-complicated. This would cover the main use-case:

```js
function * range(start, stop, step=1) {
	let i = start;
	while (step >= 0 ? i < stop : i > stop) {
		yield i;
		i += step;
	}
}
```

PR #649, opened 2017-01-17
Update link to Pino logger

jsumners:
Pino has moved into an organization. This PR updates the link to reflect that.


SamVerschueren:
Thanks!

PR #656, opened 2017-02-28
Added enduro.js

Gottwik:
hopefully up-and-coming framework to create websites with a nice admin ui extremely quickly


sindresorhus:
Thanks for the suggestion, but I'm going to have to pass on this. It's hard for me to judge a framework without any community. The project also has too low of a [bus factor](https://en.wikipedia.org/wiki/Bus_factor). Happy to reconsider some time in the future when you've gathered some traction and community around the project ;)

PR #657, opened 2017-03-04
Add mz

Rowno:
[mz](https://github.com/normalize/mz) modernizes the node.js API to current ECMAScript standards by providing promisified versions of most of node.js's methods.

This is similar to a couple of projects on the list like fs-jetpack and sander, but I think it's better because it promisifies the whole node.js API, doesn't try to change the node.js API and seems to have a lot more usage.

I'll let you decide if anything should be removed.


sindresorhus:
I think it's better to just promisify the methods yourself. It's easy enough and it works with any API, not just the core ones.

- http://bluebirdjs.com/docs/api/promise.promisifyall.html
- https://github.com/sindresorhus/pify

PR #671, opened 2017-03-10
Add pigpio

fivdi:
[pigpio](https://github.com/fivdi/pigpio) makes things possible on the Raspberry Pi that are typically not possible, for example, PWM and Servo pulses on any number of GPIOs simultaneously.


sindresorhus:
Good stuff. The list sorely needed more hardware modules :)

PR #676, opened 2017-03-16
Update/nanobox

sdomino:
Nanobox is aimed at improving Node.js development workflow. It created consistent, isolated development environments that can be deployed anywhere.

This helps maintain environment consistency across development teams and applications stages (dev, staging, production, etc.)

It's different than other projects in its class because it tackles the problem of infrastructure configuration and management as a whole, rather than just pieces of it (development environments, configuration, etc.)


sdomino:
Any needed changes or feedback before this can be merged?

florianb:
Nanobox is entirely written in Go, how does that match to a list containing "delightful Node.js packages and resources"?

I guess your PR belongs to a list like https://github.com/jondot/awesome-devenv

sdomino:
Hi, thanks for the response.

Nanobox is a resource for easily developing/deploying node.js applications. I must have misunderstood that the tools/resources here need to be written in node.js, not just something that improves node.js development.

I'm good with w/e you guys decide. I'm not trying to force something that doesn't fit with what you're trying to accomplish here. Thanks!

sindresorhus:
It doesn't have to be made in Node.js, but this is too general. We don't list Docker or Heroku either.

sdomino:
Ahh, very true. Thanks for the feedback, and your great work!

PR #680, opened 2017-03-30
Add Total.js

olenberg:
[Total.js](https://github.com/totaljs/framework) is popular (more than 2 thousand stars) framework with a set of free ready-made solutions (dashboard, blogengine, eshop, etc). I think it should be included in this list.


sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. This is a curation, so I unfortunately can't include everything.

> Total.js framework has no dependencies, so installation and upgrading is very easy.

That's the opposite of the values the Node.js community is built on. Here we build of top of existing proven work, rather than reinventing the wheel.

ORESoftware:
"That's the opposite of the values the Node.js community is built on." - who is to say that? :) I guess when you are the curator of a popular awesome-list :) There is certainly some value to having no dependencies, IMO.

PR #692, opened 2017-05-31
Add commandir to filesystem section

fvgs:
[commandir](https://github.com/fvgs/commandir) is a library offering `mkdir` and `rmdir` functions with the following features:
- Promise based
- Idempotent
- `mkdir` is recursive
- They encourage good software design and error handling by communicating exactly which directories were created or deleted, allowing for proper cleanup in the event the program aborts later on.

Other modules offer the first three features, but I am not aware of any modules offering the final feature.

I came across the need for this implementation when I noticed a program I wrote could possibly encounter an error before completing and leave an empty directory (which was going to be populated) on the user's filesystem. That's just not a nice thing to do.


fvgs:
@sindresorhus it would be great to hear your thoughts 😺

PR #699, opened 2017-06-18
new network tool: netcat.

roccomuso:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -580,6 +580,7 @@ Just type [`node.cool`](https://node.cool) to go here ✨
 - [ipify](https://github.com/sindresorhus/ipify) - Get your public IP address.
 - [getmac](https://github.com/bevry/getmac) - Get the computer MAC address.
 - [polo](https://github.com/mafintosh/polo) - Zero-config service discovery.
+- [netcat](https://github.com/roccomuso/netcat) - Fully tested Netcat port to pure js.
 
 
 ### Database


sindresorhus:
When would someone use this when working on a Node.js project? I'm genuinely curious as I've never tried original Netcat before.

roccomuso:
The module resemble the original netcat cli API, if you're familiar with netcat you'll be able to easily start with tcp sockets and udp also in Node. Using the netcat-style you're already familiar with :)

sindresorhus:
Since I've never used Netcat, it's hard for me to tell if this is something that is generally useful to most Node.js developer or if it's something just a few would use.

ORESoftware:
@roccomuso just curious why someone would want to use JS netcat instead of the standard netcat on most *nix machines?

roccomuso:
@ORESoftware The module goal is to have an easy to use and fancy API that follows netcat style. You're not going to replace netcat intended as cli utility. You could argue the same question with the [SSH2](https://github.com/mscdex/ssh2) module for example, why using it in JS when you have it on most *nix machines?

You have a full-js stack. You're not tailored to a system binary dependency and don't have to `spawn` it. You don't have to struggle with different netcat implementations and you can use it also on windows (it doesn't have netcat by default).

PR #705, opened 2017-06-28
Added v

corpix:
This is a uniform tool to manage toolchain versions. It manages only Node.js and Go versions at this time but, in theory, could be extended to manage any other language toolchain.
Goals of this tool are:
- to be portable and simple "one file" solution
- which does not require any installation or modification of bash/zsh/*rc files
- written in pure Python 3 without any other dependencies than python and git binary


sindresorhus:
Interesting idea to create something universal, but I think it's currently to immature of a project to be included here. It's not something I would recommend until it's battle-proven by many.

PR #718, opened 2017-08-10
Add timercore to miscellaneous section

fvgs:
[Timercore](https://github.com/fvgs/timercore) is a precision timer library that focuses on robustness, accuracy, and a simple API.

It is useful for applications requiring accurate timekeeping, for which there is currently no generic solution on the list.


sindresorhus:
The module looks good, but I don't think it's a good fit here. This list is targeted things that are generally useful for servers and CLIs, and I can't think of a situation when I would need this for either. Maybe you could try https://github.com/sorrycc/awesome-javascript instead ;)

PR #726, opened 2017-08-25
Add DeckTape to the command-line apps

astefanutti:
DeckTape is a PDF exporter for HTML presentation frameworks that uses Puppeteer / headless Chrome via a CLI.

It is useful to create standalone / offline versions of HTML presentations that's often require in a number of situations. It can be used to export screenshots as well.

See https://github.com/astefanutti/decktape for more information.


sindresorhus:
Thanks for the suggestion. It looks great! I'm unfortunately going to pass on this. The CLI tools list is growing too large and I need to be very picky about what I add there.

PR #730, opened 2017-09-06
Add credential-plus

simonepri:
https://github.com/simonepri/credential-plus

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -644,7 +644,7 @@ Just type [`node.cool`](https://node.cool) to go here ✨
 - [snyk](https://github.com/Snyk/snyk) - CLI and build-time tool to find & fix vulnerable npm dependencies.
 - [nsp](https://github.com/nodesecurity/nsp) - CLI tool to identify known vulnerabilities in your project.
 - [RegEx-DoS](https://github.com/jagracey/RegEx-DoS) - CLI tool to identify possible regex denial of service (ReDos) vulnerabilities in your project.
-- [credential-plus](https://github.com/simonepri/credential-plus) - Collection of secure password hashing algorithms ready to be used.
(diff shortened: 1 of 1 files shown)


sindresorhus:
Nice work, although it feels a bit awkward with a middleware API for this. I can't think of any time I would want multiple password hashing algorithms at once, and even if there are use-cases, it's not the common case. Instead, I would have done it the other way around, put common functionality in `credential-plus` and have the specific algorithms depend on it. That way I could just use the algorithm module I wanted directly without having the middleware setup dance.

simonepri:
@sindresorhus Thank you for your feedback. I really appreciate it.

The motivation between this choice was to allow users to change hash method whenever they want.
(If you need to change you hash method from one to another there is a period of time that you should support more than one hash method at once)

BTW as you pointed me out, this is not the common case.
What do you think if is the `credential-plus` package to internally require the required packages? This will remove the setup dance while keeping the modularity structure.

EG:
Instead of this:
```
const credential = require('credential-plus');
credential.install(require('credential-plus-pbkdf2'));
credential.hash('We are all humans', {func: 'pbkdf2'}).then(hash) => console.log(hash));
```

You can simply do this:
```
const credential = require('credential-plus');
credential.hash('We are all humans', {func: 'pbkdf2'}).then(hash) => console.log(hash));
```
Then `credential-plus` internally requires `credential-plus-pbkdf2` that and if not installed causes the promise to fail.

sindresorhus:
> Then credential-plus internally requires credential-plus-pbkdf2 that and if not installed causes the promise to fail.

Then you bring in a lot of unused dependencies. Maybe a hybrid approach would make sense. Let each algorithm be able to be used individually or part of `credential-plus`. That way people that only need one can use it directly or use `credential-plus` if they want the flexibility of multiple algorithms.

PR #731, opened 2017-09-06
Added Bugsnag to Logging

donpinkus:


sindresorhus:
Thanks for the suggestion, but I don't think a paid service is a good fit for this list, even if they have a free tier. https://github.com/ripienaar/free-for-dev would be a better fit.

PR #739, opened 2017-10-17
Add xlsx

reviewher:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


StarpTech:
Do you use excel viewer in Node.js :D ? This list is about backend tools.

reviewher:
Clients still pass around and prefer to deal with Excel files.  We have to process them and extract data from the server side :/

StarpTech:
My mistake when it can parse excel files in nodejs.

PR #756, opened 2017-12-07
Add yup and Prompts

erickeno:
[Yup](https://github.com/jquense/yup) is a JavaScript object schema validator and object parser which is leaner, less fancy and good bit less opinionated than joi

[prompts](https://github.com/terkelg/prompts)
**Simple**: prompts has no big dependencies nor is it broken into a dozen tiny modules that only work well together.
**User friendly**: prompt uses layout and colors to create beautiful cli interfaces.
**Promised**: uses promises and async/await. No callback hell.
**Flexible**: all prompts are independent and can be used on their own.
**Testable**: provides a way to submit answers programmatically.
**Unified**: consistent experience across all prompts.


sindresorhus:
Please read and adhere to the [contributing guidelines](../blob/master/contributing.md). Thank you 🙌

PR #769, opened 2018-01-16
Add node-jsonwebtoken

ajmalsiddiqui:
An implementation of the JSON Web Token standard for Node.
JWTs can be used in almost any scenario where stateless tokens fit in well.

Added to the miscellaneous section of packages. 

https://github.com/auth0/node-jsonwebtoken


sindresorhus:
Passing on this as people should not be using JWT: https://paragonie.com/blog/2017/03/jwt-json-web-tokens-is-bad-standard-that-everyone-should-avoid

PR #771, opened 2018-01-25
Add database-js

thiagodp:
[Database-js](https://github.com/mlaanderson/database-js) - Wrapper for multiple databases with a promise-based syntax and a JDBC-like connection.

**Database-js** is not a ODM/ORM, but a wrapper to access many databases and files with a SQL-like interface. Currently it supports Firebase, MySQL, PostgreSQL, SQLite, ActiveX Data Objects (e.g. MS Access), Excel spreadsheets, INI files, CSV files, and JSON files.


thiagodp:
@sindresorhus I suggest to add more maintainers to this project. As I type, there is `52` PRs...

SamVerschueren:
As a co-maintainer it's hard to take decisions of what should/can be added to the list and what not. The list is opinionated so adding more maintainers wouldn't really help. Sending a friendly ping once in a while can :).

PR #773, opened 2018-01-30
Add currency-symbol

atomicman57:
`currency-symbol` is a simple lightweight package that returns a country currency numeric symbol and also list all countries currency with zero dependency.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -785,6 +785,7 @@ Just type [`node.cool`](https://node.cool) to go here ✨
 
 ### Miscellaneous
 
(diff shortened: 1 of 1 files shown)


xxczaki:
> Additions should be added to the bottom of the relevant category.

😕

atomicman57:
@xxczaki I don't seem to find the appropriate category, can you help with that?

sindresorhus:
I think this would be a better fit for https://github.com/sorrycc/awesome-javascript

PR #812, opened 2018-05-22
Twofa in Auth Section

anthmarques:
Two Factor Authentication Generator for CLI.

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
Could be a useful CLI tool, but I think both the code and readme needs more work. Add a GIF or an image of how it looks like when using it. https://github.com/marionebl/svg-term-cli is good. I also wouldn't trust it with my 2FA as it just saves the services in plain text on disk...

PR #815, opened 2018-06-02
gitlink-cli - opens current repo|commit in default browser (github|gitlab|bitbucket)

bitbonsai:
https://github.com/bitbonsai/gitlink-cli

A small and super useful cli tool, that opens the default browser with the current repository page (that it gets from `git remote -v`).

## Usecase
You're in the terminal, wants to open the repository page in the browser, just type `gitlink` and it will open in the default browser. Works for github, gitlab and bitbucket, and if a partial `sha1` is provided will open the specific commit page.

If user is on a remote terminal (ssh) will just **print** the link (also available with option --print).


sindresorhus:
Thanks for the suggestion. It looks great! I'm unfortunately going to pass on this. The CLI tools list is growing too large and I need to be very picky about what I add there.

PR #828, opened 2018-07-04
Add logtown

zemd:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -230,7 +230,7 @@
 - [console-log-level](https://github.com/watson/console-log-level) - The most simple logger imaginable with support for log levels and custom prefixes.
 - [storyboard](https://github.com/guigrpa/storyboard) - End-to-end, hierarchical, real-time, colorful logs and stories.
 - [signale](https://github.com/klauscfhq/signale) - Hackable console logger with beautiful output.
-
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the suggestion, but I'm going to pass on this. I don't see enough benefit of such a facade over just using a good logger directly.

zemd:
@sindresorhus 

there are several useful cases:

1. there is no such definition "good logger" :) as we can see there are a lot good loggers.
2. one tool for both server and browser. one interface for whole application. 
3. `winston3` is better than `winston2`, but has breaking changes. with `logtown` you are changing configuration only in one place. no need to change something(imports, method arguments, method names) in any other part of app's files. 
4. easier multiple loggers integration. if you want using sentry/logdna and winston/intel - no dependency on specific implementation, but general approach. 
5. independent way of configuration for libraries - you can add `logtown` into `peerDependency` for your library and it won't mess `stdout` with unnecessary info. you can manage by yourself how to show these logs and disable if you need.

logging facades are not new, this is just one of implementation for js

 - https://stackoverflow.com/questions/3825276/whats-the-point-of-a-logging-facade
 - https://www.slf4j.org/
 - https://docs.rs/log/0.4.5/log/

PR #836, opened 2018-07-18
Add `TypeSerializer` lib

danrevah:
TypeSerializer, A library I wrote can be awesome for Node.js API's.

Article: 
https://medium.com/@danrevah/writing-awesome-node-js-api-using-decorators-in-typescript-698fc915eac5


sindresorhus:
Nothing really Node.js specific about this. Would be a better fit for https://github.com/sorrycc/awesome-javascript

PR #839, opened 2018-07-18
Add fpEs record.

johnteee:
fpEs: A small library provides simple usages of basic FP & pattern-matching/sumtype & MonadIO/Rx & Optional for Javascript

In fact I've implemented almost same features in Golang/Rust/Java/ES6. (Java is harder to release a library indeed...only golang & Rust & es6 versions are shared)

However I found ES6 is the most elegant version in them :P

Ref:

https://github.com/TeaEntityLab/fpEs
https://github.com/TeaEntityLab/fpGo
https://github.com/TeaEntityLab/fpRust

Thx :D

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
Thanks for the suggestion, but I think this would be a better fit for https://github.com/sorrycc/awesome-javascript The current FP libs are only included for convenience. I don't intend to add more.

PR #840, opened 2018-07-20
Added create-soda-app

igorkamyshev:
Hello! I think, https://github.com/solid-soda/create-soda-app good candidate for this list. It's not framework, but it simple way to create app with Express, TypeScript and so on.

Please, give me your feedback.


igorkamyshev:
Ok. I understand.

PR #858, opened 2018-09-09
Add wetland to the list of awesome.

RWOverdijk:
Nice list, I just noticed my ORM is missing, so the list wasn't truly awesome yet. :D

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -586,6 +586,7 @@
 	- [pg-promise](https://github.com/vitaly-t/pg-promise) - PostgreSQL framework for native SQL using promises.
 	- [Objection.js](https://github.com/Vincit/objection.js) - Lightweight ORM built on the SQL query builder Knex.
 	- [TypeORM](https://github.com/typeorm/typeorm) - ORM for PostgreSQL, MariaDB, MySQL, SQLite, and more.
+	- [Wetland](https://github.com/spoonx/wetland) - JPA-spec based ORM. Supports PostgreSQL, MySQL, MSSQL, SQLite3 and more.
 - Query builder
 	- [Knex](https://github.com/tgriesser/knex) - Query builder for PostgreSQL, MySQL and SQLite3, designed to be flexible, portable, and fun to use.
 - Other


sindresorhus:
There are already a bunch of ORMs in this list. How is your better/different from them?

RWOverdijk:
Because it's based on the JPA-spec and has a nice migrator. The way it combines strict migration-based database design and dev migrations for ease of development is what I like most about it. It has proven itself to be stable and has been used in production for a while now.

I'm not entirely sure what you want me to say, it's a curated list. 😅

sindresorhus:
You mention JPA-spec, but the Wetland readme and docs have no mention of it. At least from my quick search). You should at minimum mention it in the project readme with a summary of what it is and its benefits.

RWOverdijk:
@sindresorhus There's a list of features in the readme and a pretty extensive tutorial in the documentation to help explain why it's awesome. The tutorial doesn't take that long and shows you almost all features. https://wetland.spoonx.org/Tutorial/setting-up.html

The reason I don't document the fact that it's based on the JPA spec is because it doesn't follow it to the letter (because it can't, the platform is different). However, all the patterns are there and the implementation is closely modelled to it. I mentioned it here to help you understand why I think it's a good ORM quality wise, and why I think it's different and deserves to be listed as a quality ORM.

Having alternatives doesn't hurt. It's a battle-tested, well maintained ORM that's being used in production.

RWOverdijk:
> That's all fine. What I'm having a problem with is that you're including JPA-spec based ORM. in the description, but nowhere do you explain to the user what that means. That makes the part of the description totally useless.

You're right. I'll add a description about that.

RWOverdijk:
@sindresorhus I have updated the landing of the docs and the readme. I've also added a link, and the index of the specification for those familiar with the patterns used by it will add extra context.

RWOverdijk:
Nothing? :) We made [another piece of awesome](http://stixjs.io) for nodejs by now, should I bother making a pull request for that?

RWOverdijk:
Cool. Thanks for replying.

PR #863, opened 2018-09-28
Add 'MagiCLI' lib

DiegoZoracKy:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

The goal of [MagiCLI](https://github.com/DiegoZoracKy/magicli) is to automagically generate a command-line interface for any module, without the need to create anything by hand. Just `require('magicli')();` and your module is ready to be executed via CLI. It is also capable of executing any module via CLI, even if the target module doesn't provide one. Just install it globally (or use *npx*) and: `$ magicli ./path/to-a-module-file-or-package.js --help`.

Articles written about it:

[Introducing MagiCLI: Automagically generates a command-line interface (CLI) for any module
](https://hackernoon.com/introducing-magicli-automagically-generates-a-command-line-interface-cli-for-any-module-49543e50f86d)

[Executing any Node.js module or .js file via CLI without creating any command-line interface at all
](https://hackernoon.com/executing-any-node-js-module-or-js-file-via-cli-without-creating-any-command-line-interface-at-all-6f3a125fdf5a)

1_3mw1jcvfq2ijhxmnd6di9q

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -284,6 +284,7 @@
 - [gradient-string](https://github.com/bokub/gradient-string) - Beautiful color gradients in terminal output.
 - [oclif](https://github.com/oclif/oclif) - CLI framework complete with parser, automatic documentation, testing, and plugins.
 - [term-size](https://github.com/sindresorhus/term-size) - Reliably get the terminal window size.
+- [magicli](https://github.com/DiegoZoracKy/magicli) - Automagically generates command-line interfaces for any module, or, execute any module via CLI.
 
 
 ### Build tools


DiegoZoracKy:
Conflicts resolved! Now it is ready to be merged ;)

sindresorhus:
It's a very cool experiment, but I don't think it's production-ready enough to include here right now. There's no sign of anyone using it. I think it needs to be proven in production first.

DiegoZoracKy:
Hi @sindresorhus, how do you assess if there is someone using a module? That's a genuine question. I tend to look at downloads stats over time, and based on that it seems that MagiCLI is being used a lot. At this case, it is not reliable to check if it is included as a dependency on other modules, as it can be heavily used as a standalone CLI solution. The issue here is that although its use is increasing, and we can see that people are using more and more, it is not known who is using it and to what purpose.

PR #866, opened 2018-10-06
Add Purdy - A useful pretty printer for debugging

danielb2:
Pretty print objects in real purdy colors. Allows clearer visualization of objects than you get from most pretty printers due to colors. It will also print out the complete path to an object, something that's extremely useful for debugging. Purdy will also print out the path to access a variable using Hoek/_.get format making it useful on accessing values.


sindresorhus:
Thanks for the suggestion. I'm unfortunately going to pass on this one. I have a lot of PRs to review, so I'm not able to write a custom response.

danielb2:
🙄

PR #876, opened 2018-11-06
Add sade to Command-line utilities

radum:


sindresorhus:
I don't really see any benefit of this over https://github.com/yargs/yargs

radum:
Functional, yes I agree, I only added it as an alternative comparing the size of each one of them.

PR #886, opened 2018-12-27
Add a French book about Node.js

oncletom:
Currently ranked in 15th position Amazon.fr, for programming books.

I'm a bit biased, because I wrote it.


sindresorhus:
Thanks for the suggestion, but I'm not accepting non-English resources as I wouldn't be able to judge their quality.

PR #892, opened 2019-01-04
Add Refo to the Build tools

kireerik:
https://github.com/kireerik/refo

It is simpler and easier to customize than other build tools.


sindresorhus:

PR #908, opened 2019-03-05
Add manipula

litichevskiydv:
Submitted library natively supports iterable protocol and provides functional for grouping data by complex keys, ordering iterables by multiple criteria and something else. Moreover basic operations with sets, represented by iterables, such as intersection, subtracting, union, etc. are also supported.  Results are evaluated lazily.

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
Thanks for the suggestion. I'm unfortunately going to pass on this one. It looks like a good start, but I don't think it's mature enough to be added here. It also lacks an audience. Get people to use your library first and prove it's awesome, then add it here. Not the other way around.

PR #936, opened 2019-05-17
Add play-sound

gakada:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

https://github.com/shime/play-sound

It doesn't provide fallback binaries (probably because those can be pretty bulky), but otherwise it is similar to [clipboardy](https://github.com/sindresorhus/clipboardy), allowing for a simple API, e.g.

```js
play('some.mp3')
```

which will attempt to play it by `spawn`ing an available system player.


sindresorhus:

PR #943, opened 2019-06-13
Add ziffer

dresende:
Format a number in any representation (any currency, thousand separator, arabic, etc). Also, do the opposite and extract a number from a formatted string.

https://github.com/dresende/node-ziffer

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
https://github.com/sindresorhus/awesome-nodejs/blob/master/contributing.md

dresende:
Could you be more specific?

sindresorhus:
I'm going to pass on this as users are better off using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat or a polyfill if they need to support older browsers.

PR #947, opened 2019-06-19
Add better-queue

AxelTerizaki:
I looked through different job queue systems for my standalone, desktop app, and found this one, which does the job (so to speak) quite well, and it doesn't need me to install/embed Redis. It's simple and efficient and perfect for small workloads when you need a queue system (for example to manage a download queue)


AxelTerizaki:
Github link : https://github.com/diamondio/better-queue

PR #949, opened 2019-06-19
Add qrcode

AxelTerizaki:
Useful QRCode generator which can be used as CLI tool, module (both for node and browser)


AxelTerizaki:
Forgot to add the repo link in the PR, my bad : https://github.com/soldair/node-qrcode

PR #950, opened 2019-06-19
Add node-id3

AxelTerizaki:
Pretty useful module to read/write id3 tags from mp3 files. Used it to detect if a mp3 has a cover image embedded before feeding it to a music player.


AxelTerizaki:
And I once again forgot the github link : https://github.com/Zazama/node-id3

sindresorhus:
Thanks for the suggestion, but I don't think this is generally useful enough. I'm trying to contain the list as it's growing too large.

PR #951, opened 2019-06-20
Add thetool

ak239:
Github link: https://github.com/sfninja/thetool

> thetool is a CLI tool to capture different cpu, memory and other profiles for your node app in Chrome DevTools friendly format.

Please check [readme.md](https://github.com/sfninja/thetool#readme) for more details.


sindresorhus:
How is this different from just using Chrome DevTools directly with Node.js?

ak239:
It connects to child processes and workers, `ndb` supports only child processes but does not support workers. Stock DevTools does not support child processes.
It is small and can be used in other environments, e.g. user can run app with `thetool` on server and send report back to his machine for further analysis.
It supports more formats than DevTools, e.g. coverage data.

sindresorhus:
That makes sense. I would recommend opening a PR on the thetool repo to improve its readme with the above info. I cannot be the only one wondering "what's the point".

PR #953, opened 2019-06-21
Add JKT and PodengJs

slaveofcode:
JKT is the one of my project to resolve data structure consistency among our JSON values, and also JKT is acting as a parser on RESTful api response at [usetada.com](https://usetada.com). That library is simply and fast enough to make an object using a string literal (you can even create with one line only), which is what my team needs in order to easy-setup rules on our json objects inside the application.

And then the biggest part is coming, the project on some feature needs a more way to parse with more advanced feature, which is unsupported by JKT. Based on the first problem (which solved by jkt) then I built PodengJs which provide the json parser but also validator (optional, by config). That's why I also put the PodengJs on validator section.

Please just hit me if the purpose of PodengJs here is not matching the criteria of these validator list, I'll just remove and create an update.

Github: 
- https://github.com/slaveofcode/podeng
- https://github.com/slaveofcode/jkt

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -501,6 +501,7 @@
 - [property-validator](https://github.com/nettofarah/property-validator) - Easy property validation for Express.
 - [schema-inspector](https://github.com/Atinux/schema-inspector) - JSON API sanitization and validation.
 - [ajv](https://github.com/epoberezkin/ajv) - The fastest JSON Schema validator. Supports v5 proposals.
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the suggestion. I'm unfortunately going to pass on this one. I have a lot of PRs to review, so I'm not able to write a custom response.

PR #955, opened 2019-07-10
Financial category and first suggestion added

bobbilee19:
Realized I wasn't seeing any payment/financial packages so I suggested one SDK for taking payments.


sindresorhus:
Thanks, but I don't really want to list all kinds of SDKs for companies. This list is mostly independent projects.

PR #957, opened 2019-07-13
Add next-boilerplate

omeraplak:
This boilerplate make it easier to get started with a well-structured Next.js and TypeScript application.
By the end of setup, you'll have a Next.js project and features which is specified at the below.

https://pankod.github.io/next-boilerplate/

https://github.com/pankod/next-boilerplate/

added under #web-frameworks


sindresorhus:
While Next.js does do server rendering, I think this is a better fit for https://github.com/sorrycc/awesome-javascript

PR #966, opened 2019-08-11
Add Segment as sponsor

vadimdemedes:
This is how it looks:

CleanShot 2019-08-11 at 11 03 14@2x

Thanks!


sindresorhus:
Here's the SVG logo. Can you use it instead?

[segment-logo.svg.zip](https://github.com/sindresorhus/awesome-nodejs/files/3490545/segment-logo.svg.zip)

(Found it on the Segment website and just minified it)

vadimdemedes:
Updated! I force pushed so that jpg logo doesn't get stored in git history.

PR #971, opened 2019-09-17
Add nanoid

florianb:
https://github.com/ai/nanoid is an actively maintained id-generator which compares best to UUIDv4 (random-based id generation) which generates ids with a similar collision probability, but less size (21 symbols vs 36) and ~16% faster. It has security in mind (using crypts random number generator and providing a company-backed disclosure procedure.

It is very helpful at any point heavily relying on collision-free ids, like f.e. cassandra or couchdb document-ids.

I don't know if mad science is the best category.

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
It's not really mad science, but definitely useful.

florianb:
You're right, i'd propose then **Text** or **Misc** (even though the latter is already pretty stuffed).

sindresorhus:
`Text` then

florianb:
Aye :vulcan_salute:

PR #972, opened 2019-09-17
Add modern-node

sheerun:
https://github.com/sheerun/modern-node

> For CLI tools, the bar is especially high, and unless it's something very awesome

I think it's awesome through the roof but it's up to you to decide :P

This project is similar just to your boilerplate but works differently:

1. It's not meant for cloning but installing or creating with `create-modern-node`
2. It uses jest instead of ava, it uses prettier+eslint+standard instead of xo
3. It has built in precommit hook so installing husky is not necessary

I've put it next to your package instead of at the bottom because these two are most similar in functionality.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -921,7 +921,7 @@
 
 - [nodebots](http://nodebots.io) - Robots powered by JavaScript.
 - [node-module-boilerplate](https://github.com/sindresorhus/node-module-boilerplate) - Boilerplate to kickstart creating a node module.
-- [modern-node](https://github.com/sheerun/modern-node) - Toolkit for creating node modules with Jest+Prettier+Eslint+Standard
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think the naming is unfortunate. It has the same problem as Standard. The naming fools people into thinking this is the "modern" way of doing node modules, while it's really just your opinionated preference, just like Standard.

sindresorhus:
I mean, I’m not saying I won’t accept it, but maybe you want to do something about the marketing?

sheerun:
Unfortunately I cannot change name of module, but I've changed marketing to "Toolkit for creating node modules with Jest testing, Prettier formatting, and Standard linting" for now

sheerun:
I've updated description here as well

PR #977, opened 2019-10-06
Added NeDB

archiebaer:
NeDB is a lightweight and self-contained database that's fast and has an API that is a subset of MongoDB's.


crazy4groovy:
Looks pretty old. Is it stable?

archiebaer:
I use it for a lot of projects, but obviously, it's not designed for large-scale stuff. The only time I've had anything using it crash on me was when I was building a discord bot that received a few hundred thousand messages in the 24 hours before it crashed.

sindresorhus:
It's already in the list.

archiebaer:
Ah, it wasn't when I made the PR.

PR #980, opened 2019-10-13
added new package: request-promise: Request library with promises support.

AkhilMukkamala:
Why it should be added ?

For beginners, they can find request module in the list, But if they want the same request module with builtin promises support, then this is the package they should check for.

Link to the package: https://github.com/request/request-promise


sindresorhus:
Request is no longer actively maintained, so doesn’t make sense to add wrappers for it.

AkhilMukkamala:
Thanks for letting me know. I've still using request module. @sindresorhus

PR #991, opened 2019-11-01
Add runtime-configurator

damlys:
`runtime-configurator` package allows to fill in runtime configuration with data from files, env vars and cli args. It validates provided values, generates documentation which might be displayed in terminal, does not collide with other CLI packages, is projected to define config in declarative style and does not impact the application's performance

Currently it is used in three enterprise/commercial applications and has stable API


papb:
Why did you close this?

damlys:
Too long without response

papb:
I think you should leave it open... One day someone might look at it :)

PR #1004, opened 2020-01-03
added dree

euberdeveloper:
[github](https://github.com/euberdeveloper/dree)
[npm](https://www.npmjs.com/package/dree)

Even if it has not lots of downloads this is the most complete and powerful module about a directory tree. I cannot find a similar module with more options of dree. It could be useful for instance when making the string representation of a project and putting it in a readme. Having also callbacks, I used this module also for a JavaScript template compiler. I used it to walk the directory, filtering only the files I wanted and having lots of information each of those file as a callback parameter. It can also be used with testing: I needed to test if a function produced a directory with a certain structure and I used this module to check if the structure was the one expected.

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:

PR #1026, opened 2020-02-29
Add jsep tiny expression parser package

bertyhell:
This package is a lot more lightweight than a full AST parser since it only parses a single line expression, so it is easier for users to get started with.

I've used it in a project and it was a breeze to setup. The documentation is also really cool, since its just an annotated version of the source code,and still it is very clear and descriptive. 

https://github.com/soney/jsep

Yes, and you should add some text to this that says to add a checked checkbox verifying the guidelines were read, then you can make a github bot that closes any submissions without this checkbox in the description? :p

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -712,6 +712,7 @@
 - [Acorn](https://github.com/ternjs/acorn) - Tiny, fast JavaScript parser.
 - [babel-parser](https://github.com/babel/babel/tree/master/packages/babel-parser) - JavaScript parser used in Babel.
 - [cherow](https://github.com/cherow/cherow) - JavaScript parser with focus on performance and stability.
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the suggestion, I'm however going to pass on this for two reasons.

1. I don't think it's generally useful enough. It's not something most would need. And those who do can just search for it.
2. It has not had a release in more than 2 years and generally doesn't seem very well maintained.

sindresorhus:
> then you can make a github bot that closes any submissions without this checkbox in the description? :p

I've considered setting up a bot, yes, but in the end, that will take more time than just closing it myself.

PR #1036, opened 2020-04-23
docs: ✨ add imgix-core-js as image library

frederickfogerty:
This PR adds imgix-core-js to the list of awesome image components. This is an awesome library because it makes images rendered by it automatically more performant, and responsive via srcsets! The repository has been around for 5 years, well-documented, and is commercially-supported.

You can check out the package at the [GH repo](https://github.com/imgix/imgix-core-js), [NPM](https://www.npmjs.com/package/imgix-core-js), or at [our official documentation](https://docs.imgix.com/libraries/imgix-core-js).


sindresorhus:
Thanks for the suggesting, but I prefer to not include libraries that are SDKs or require a closed-source server.

PR #1039, opened 2020-06-07
Suggested addition of Imageflow

anshulrgoyal:
Imageflow is open-source software that scales, edits, and optimizes images. It is faster than any module available in this space and supports the batch operation. It aims to put security first and yet be fast.
Features:-
- Support GIF,webP , PNG, JPEG.
- Resize, crop, watermark, add a border, etc


sindresorhus:
Thanks for your suggestion, but I think this package is too immature. It's at version 0.0.1. Happy to reconsider when it's at 1.0.0 and widely used in production.

PR #1042, opened 2020-06-26
Add rubico

richytong:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

rubico is a robust, highly optimized syntax for async agnostic functional programming in JavaScript. async agnostic for rubico means you can pass in synchronous functions or asynchronous functions to rubico methods with reasonable behavior. For example, rubico's `map` will return a synchronous value when used with the synchronous `square`, but returns a Promise when used with the asynchronous `asyncSquare`

```javascript
const square = x => x ** 2

const asyncSquare = x => Promise.resolve(x ** 2)

const { map } = rubico

map(
  square,
)([1, 2, 3, 4, 5]) // > [1, 4, 9, 16, 25]

map(
  asyncSquare,
)([1, 2, 3, 4, 5]) // > Promise { 1, 4, 9, 16, 25 }
```

for rubico, performance and interface are king. rubico is highly expressive, yet takes care that it does not leak memory. rubico does not have any physical dependencies, but the more complex asynchronous algorithms are heavily advised by stackoverflow. For example, the `map.pool` implementation is heavily advised by [this question](https://stackoverflow.com/questions/62037349/rubicos-map-pool-array-implementation)

Thank you for your consideration.

https://github.com/a-synchronous/rubico/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -207,6 +207,7 @@
 - [RxJS](https://github.com/reactivex/rxjs) - Functional reactive library for transforming, composing, and querying various kinds of data.
 - [Lazy.js](https://github.com/dtao/lazy.js) - Utility library similar to lodash/Underscore but with lazy evaluation, which can translate to superior performance in many cases.
 - [Kefir.js](https://github.com/kefirjs/kefir) - Reactive library with focus on high performance and low memory usage.
+- [rubico](https://github.com/a-synchronous/rubico) - Async agnostic functional programming.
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the suggestion, but I think this might be a better fit for https://github.com/sorrycc/awesome-javascript

The functional list here is just a small selection of the most popular ones as it's not directly related to Node.js.

PR #1055, opened 2020-09-05
Update description for Grant package

simov:
:wave: I decided to update the description of my OAuth client, that I initially added to this list in 2015 https://github.com/sindresorhus/awesome-nodejs/pull/230

A lot have changed since then, and so I though that it might be useful to outline the key selling points as of today, because that might be useful information for anyone searching for such solution.

Main difference is that [Grant](https://github.com/simov/grant) now ships with 4 serverless handlers, making it a viable solution for Lambda Functions, and then I also added a handler for Fastify.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -632,7 +632,7 @@
 ### Authentication
 
 - [Passport](https://github.com/jaredhanson/passport) - Simple, unobtrusive authentication.
-- [Grant](https://github.com/simov/grant) - 200+ OAuth providers for Express, Koa, Hapi, Fastify, AWS Lambda, Azure, Google Cloud, Vercel.
+- [Grant](https://github.com/simov/grant) - OAuth providers for Express, Koa, Hapi, Fastify, AWS Lambda, Azure, Google Cloud, Vercel, and many more.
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -632,7 +632,7 @@
 ### Authentication
 
 - [Passport](https://github.com/jaredhanson/passport) - Simple, unobtrusive authentication.
-- [Grant](https://github.com/simov/grant) - OAuth middleware for Express, Koa, and Hapi.
+- [Grant](https://github.com/simov/grant) - 200+ OAuth providers for Express, Koa, Hapi, Fastify, AWS Lambda, Azure, Google Cloud, Vercel.
```
sindresorhus, line 635: I don't want to have any numbers as it will have to be kept up to date.
```suggestion
- [Grant](https://github.com/simov/grant) - OAuth providers for Express, Koa, Hapi, Fastify, AWS Lambda, Azure, Google Cloud, Vercel, and many more.
```

simov:
Sounds good, thanks @sindresorhus :+1:

PR #1057, opened 2020-09-10
Add Graceful Shutdown with NodeJS and Terminus

wakeupmh:
Hi,
I'm Marcos Henrique and I'm a software engineer at Dasa
I though this small **article** about graceful shutdown might be a good contribution to your list.

Thanks for considering it!


sindresorhus:
The article looks good, however, GoDaddy is [a shitty company](https://karveldigital.com/why-i-dont-use-godaddy-you-shouldnt-either/) and I don't want to link to anything related to them.

I'm happy to consider another article written for a different process manager like PM2.

PR #1058, opened 2020-09-10
Add to Misc - express decorator router

wakeupmh:
### Looking for that package that will help you test your projects by decoupling intelligently?

**Your problems end here and now**, I present to you the package whose premise **doesn't need transpilation**, and **can be used with koa, express or any package following express API**.

You who want to escape from transpilation and are a hipster who likes the good old vanilla JS.

This package is intended to avoid unnecessary creation of two files where one contains the route definition and the other has the function that handles the route request/response processes, leaving simpler maintenance and more scalable code.

With **only 32kb** minified (source: [bundlephobia](https://bundlephobia.com/result?p=express-decorator-router@0.1.1)), being able to inject the dependencies in a "manual" way or with the extremely tough tool [awilix](https://github.com/jeffijoe/awilix)


olavoparno:
Wow dude this lib helped me alot back in the day!!!

sindresorhus:
Thanks for the PR. However, this would be a better fit for https://github.com/rajikaimal/awesome-express I don't want to end up with too many projects specific plugins/stuff here.

PR #1060, opened 2020-09-10
add LIVR to data validation

wakeupmh:
LIVR - Is a ightweight JavaScript validator supporting Language Independent Validation Rules Specification


sindresorhus:
Thanks for the suggestion, however, the project doesn't have any changelog. You also included a lot of unrelated changes.

PR #1073, opened 2020-10-26
Add editly

mifi:
https://github.com/mifi/editly

Slick, declarative video editing API. Lets you easily create video clips using code


sindresorhus:
Nice!

PR #1081, opened 2020-11-24
Add Superstruct

ianstormtaylor:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

Name: `superstruct`
GitHub: https://github.com/ianstormtaylor/superstruct
Yarn: https://yarnpkg.com/package/superstruct
NPM: https://www.npmjs.com/package/superstruct
Stars: 4.6k
Downloads: 733.3k / mo

Superstruct makes it easy to define interfaces and then validate JavaScript data against them. Its type annotation API was inspired by Typescript, Flow, Go, and GraphQL, giving it a familiar and easy to understand API.

But Superstruct is designed for validating data at runtime, so it throws (or returns) detailed runtime errors for you or your end users. This is especially useful in situations like accepting arbitrary input in a REST or GraphQL API. But it can even be used to validate internal data structures at runtime when needed.

```ts
import { assert, object, number, string, array } from 'superstruct'

const Article = object({
  id: number(),
  title: string(),
  tags: array(string()),
  author: object({
    id: number(),
  }),
})

const data = {
  id: 34,
  title: 'Hello World',
  tags: ['news', 'features'],
  author: {
    id: 1,
  },
}

assert(data, Article)
// This will throw an error when the data is invalid.
// If you'd rather not throw, you can use `is()` or `validate()`.
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -506,6 +506,7 @@
 - [property-validator](https://github.com/nettofarah/property-validator) - Easy property validation for Express.
 - [schema-inspector](https://github.com/Atinux/schema-inspector) - JSON API sanitization and validation.
 - [ajv](https://github.com/epoberezkin/ajv) - The fastest JSON Schema validator. Supports v5, v6 and v7 proposals.
+- [Superstruct](https://github.com/ianstormtaylor/superstruct) - Simple and composable way to validate data in JavaScript (and TypeScript).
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I've used this in the past. It's great. 👍🏻

PR #1085, opened 2020-12-09
Add istanbul badges readme

olavoparno:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

github: https://github.com/olavoparno/istanbul-badges-readme


sindresorhus:
Thanks for the suggestion, but I think this is too Istanbul specific. This list is more about general tools than plugins/packages for specific projects. I would rather try to get it listed in the Istanbul readme.

PR #1091, opened 2020-12-29
Add Rollpkg to Build Tools

rafgraph:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

Rollpkg - https://github.com/rafgraph/rollpkg
Example package setup to use Rollpkg - https://github.com/rafgraph/rollpkg-example-package

Rollpkg is a zero config build tool to create packages with Rollup and TypeScript. It creates `esm`, `cjs` and `umd` builds for development and production, and fully supports tree shaking. Default configs are also provided for TypeScript, Prettier, ESLint, and Jest for a complete zero decision setup.


sindresorhus:
Thanks for the suggestion. I'm unfortunately going to pass on this. There are just so many tools like this and it's hard for me to pick which ones to include as they're all very opinionated in different ways. I also don't particularly agree with the solution here. I personally think packages should not use bundling (transpiling TS is fine though). You should also not use the `main` and `module` fields, but rather `type` and `exports` as those are standards.

PR #1094, opened 2021-01-14
Update readme.md

davidnussio:
I want to share powerful collection of Algebraic Data Types and FP utils


sindresorhus:
I think this is a better fit for https://github.com/sorrycc/awesome-javascript

PR #1107, opened 2021-03-13
Add Ts.ED

Romakita:
Hello @sindresorhus 
Thanks for this awesome list :)

I propose to add the [Ts.ED](https://tsed.io) framework. Ts.ED help developers to build a Rest/Web application based on Express.js or Koa.js by using TypeScript decorators and common patterns. The framework is pretty simple to use, support many other frameworks/library/third party and his learning curve is fast.

The documentation is pretty complete and provide many examples to help developers to be ready to dev.

His goal is to support other web framework like fastify which allows to have a "cross http frameworks" framework (koa, express, fastify).

Thank you in advance for taking the time to evaluate my proposal,
See you and thanks again for your awesome work.
Romain


sindresorhus:
Looks like a useful framework. I think it would benefit from a better name though. `Ts.ED` is hard to remember and looks weird. Just my two cents.

Romakita:
Thanks a lot @sindresorhus. I totally agree, I totally missed the name. I don't know how complicated changing the name of a framework is at this point.

I will try to work on this aspect and plan a change in this direction thank you for your feedback ;)

Thank you for the pr and for your huge work!
Romain

PR #1137, opened 2021-08-25
Include sonic < express > in documentation section 

Tiemma:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

Link to the package: https://github.com/Tiemma/sonic-express

Reasons for including: It adds support for API documentation on the fly which is not common amongst the Express ecosystem.

It has some similar design considerations to [Flask RestX](https://github.com/python-restx/flask-restx) but due to Node's limitation in support for implicitly documenting APIs, this uses tests or gives the user the ability to generate as they work on the APIs themselves (self-documented at the end of the day). 

As far as requests hit the API, the middle ware would document them as they go and overwrite subsequent requests based on the status code and other param.

I currently built this and it is used at [Replex](https://replex.io) where I work and over the months, I have on-boarded users who've had similar problems and issued fixes and API updates and they use it too hassle free.

Looking to extend visibility for it as it works really well from the feedback I've gotten and ease of integration from external parties.


sindresorhus:
Looks like a useful package, but I think this would be a better fit for: https://github.com/rajikaimal/awesome-express I'm trying not to include too many framework-specific packages as there's just so many.

PR #1140, opened 2021-09-02
Add express-actuator

rcruzper:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

Express-actuator is a middleware which creates a series of endpoints to help you monitor and manage your application when it's pushed to production. It's useful when you run your applications on container based orchestrators, as Kubernetes, and you are in need of endpoints for readiness/liveness probe.


sindresorhus:
Thanks for the suggestion. I think this would be a better fit for: https://github.com/rajikaimal/awesome-express I'm trying not to include too many framework-specific packages as there's just so many.

PR #1149, opened 2021-10-15
feat: add petitio

tbnritzdoge:
https://github.com/helperdiscord/petitio

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
I have been doing HTTP clients for years. My most important lesson is that tiny HTTP clients are the ones no one are using, because when a HTTP client meets reality of networking, it's impossible to keep it small. Also, microbenchmarking HTTP clients with a local server makes no sense. The bottleneck of HTTP clients is networking.

I'm going to pass on this for now, but I'll reconsider if your HTTP client becomes widely used and more mature.

PR #1150, opened 2021-10-20
add proxyquire

wantao666:


sindresorhus:
I have mixed feelings about adding it. On one hand, it has been quite useful in the past. However, it has been inactive for a couple of years, and it does not support ESM (which is not possible yet). ESM support would require a complete rewrite and I don't think that will happen there.

PR #1167, opened 2022-03-01
add zx 

xiaodang:
zx is a new tool from Google that lets you write Bash-like shell scripts with JavaScript.


sindresorhus:
From their readme:

> Disclaimer: This is not an officially supported Google product.

PR #1172, opened 2022-03-12
Validate that links are up (once a week)

simeg:
Every Sunday at midnight this Github action will validate that all
http/https links responds with 200. If a link has been permanently
removed (or just moved) it can be removed/changed which leads to a
higher quality list.

This check is already in place in the mac-setup[0] repo and is working
well. It won't prevent any PRs from getting merged, and optionally a
badge can be added to the README to highlight the link status (see
mac-setup repo).

[0]: https://github.com/sb2nov/mac-setup/

---

Hi @sindresorhus! I'm curious what you think of my suggestion. I think this will increase the quality of this nice repo! 🙂 It will make commits like [this one](https://github.com/sindresorhus/awesome-nodejs/commit/041a511ee4ab86b2069a2071d1e89abdc38a040d) easier to create as the workflow will figure out what URLs are != 200.

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -1,20 +1,15 @@
 name: Links
-
 on:
   schedule:
     # Run at 00:00 UTC every Sunday
     - cron: '0 0 * * */7'
-
 env:
   FORCE_COLOR: 1
-
 jobs:
   build:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v2
-
       - name: Validate that links are up
         uses: simeg/urlsup-action@v1.0.0
         with:
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -1,0 +1,21 @@
+name: Links
+
+on:
+  schedule:
+    # Run at 00:00 UTC every Sunday
+    - cron: '0 0 * * */7'
+
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Validate that links are up
+        uses: simeg/urlsup-action@v1.0.0
+        with:
+          args: readme.md --threads 10


.github/workflows/links.yml
```diff
@@ -0,0 +1,36 @@
...
+          -not -path '*/\.git/*' -not -path '*/\.github/*' \
+          -name "*.md")
+
+          echo $files_to_check >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+
+      - name: Validate that links are up
+        uses: simeg/urlsup-action@v1.0.0
+        with:
+          args: ${{ env.FILES_TO_CHECK }} --threads 10
```
sindresorhus, line 36: I don't see the point of using environment variables. The flags can just be passed directly here.

.github/workflows/links.yml
```diff
@@ -0,0 +1,36 @@
...
+  FORCE_COLOR: 1
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Find markdown files
```
sindresorhus, line 18:
```suggestion
      - name: Find Markdown files
```

.github/workflows/links.yml
simeg: Assuming `readme.md` is the only file we want to verify then yeah, we can just pass it

simeg:
Note that on the first run it may trigger some 429s and other things that will need tweaking.

simeg:
@sindresorhus You need to [activate Github actions](https://docs.github.com/en/enterprise-server@3.1/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository), or allow this action to be run. Or perhaps it's active just that I'm not seeing the results?

sindresorhus:
Yeah, I have activated actions now. It will run next Sunday then.

sindresorhus:
@simeg It's unclear to me why it's failing now: https://github.com/sindresorhus/awesome-nodejs/runs/6243287642?check_suite_focus=true

simeg:
@sindresorhus Github.com is rate limiting it, hence the `429` response codes. For now we can ignore `429` codes if you'd like. I think it's solvable by passing a `GITHUB_TOKEN´, just gotta find some time to implement it.

PR #1180, opened 2022-07-19
add @egomobile/http-server in "Web frameworks"

mkloubertego:
# @egomobile/http-server

This is a module that is very similar to Express.js and offers most of its known API.

Apart from the well-known methods, such as [get()](https://egomobile.github.io/node-http-server/interfaces/IHttpServer.html#get) or [post()](https://egomobile.github.io/node-http-server/interfaces/IHttpServer.html#post), well-known middlewares, such as [cors()](https://github.com/expressjs/cors), can also be included.

```typescript
import cors from "cors";  // npm i && npm i -D @types/cors
import createServer from "@egomobile/http-server";

const app = createServer();

// use with "cors" middleware
app.get("/", [ cors() ], async (request, response) => {
  // all handlers are Promise-optimized, so you do not need a
  // response.end() anymore
});

app.post("/", [ cors() ], async (request, response) => {
  // do your POST here
});

await app.listen(8080);
```

In addition, a lot of useful [middlewares](https://github.com/egomobile/node-http-server/wiki/Middlewares#build-in-middlewares) are already built in.

```typescript
import createServer, { json }  from "@egomobile/http-server";

// ...

app.post("/", [ json() ], async (request, response) => {
  // now you should have a parsed and valid
  // JavaScript object in request.body
  const { body } = request;
});

// ...
```

The server is especially optimized for the use in the microservice area, but can also be used like Express for monolith backends or websites.

The library is also characterized by the fact that you can implement your endpoints extremely fast with the help of [decorators](https://github.com/egomobile/node-http-server/wiki/Controllers) and it is also possible to realize schema validations in the simplest way with the help of [Joi](https://joi.dev/) or to provide API documentation as [Swagger UI](https://github.com/swagger-api/swagger-ui).

```typescript
// path inside project folder: /controllers/@abc/index.ts

import {
  Controller,
  ControllerBase,
  DELETE,
  GET,
  POST,
  IHttpRequest,
  IHttpResponse,
} from "@egomobile/http-server";

// base path: /:abc
//            each @ prefix in a directory's base name will be converted to a parameter
//            which can be accessed via 'params' prop of a 'request' context instance
//            of a method
@Controller()
export default class WithParamsController extends ControllerBase {
  @GET() // full path is: /:abc
  //
  // 'index' as name of the method will also be handled an empty string
  // and we do not specify an explicit (sub)path in '@GET()'
  async index(request: IHttpRequest, response: IHttpResponse) {
    response.write("value of 'abc' is: " + request.params!.abc);
  }

  @POST() // full path is: /:abc/foo
  //
  // the name of method ('foo') is different to 'index'
  // and we do not specify an explicit (sub)path in '@POST()'
  async foo(request: IHttpRequest, response: IHttpResponse) {
    response.write("value of 'abc' is: " + request.params!.abc);
  }

  @DELETE("/baz/:buzz") // full path is: /:abc/baz/:buzz
  //
  // the name of method ('bar') is ignored here
  // because we specified an explicit (sub)path in '@DELETE()'
  //
  // this will also happen, even if the method name is 'index'
  async bar(request: IHttpRequest, response: IHttpResponse) {
    response.write("value of 'abc' is: " + request.params!.abc);
    response.write("value of 'buzz' is: " + request.params!.buzz);
  }
}
```

An instance is [between 3-4 faster than Express](https://github.com/egomobile/node-http-server#benchmarks-) because, among other things, it dispenses with most of its auxiliary functions to achieve optimum speed.

A full overview provides the [README](https://github.com/egomobile/node-http-server#readme) and the [projects's wiki](https://github.com/egomobile/node-http-server/wiki)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -492,6 +492,7 @@
 - [Marble.js](https://github.com/marblejs/marble) - Functional reactive framework for building server-side apps, based on TypeScript and RxJS.
 - [Lad](https://github.com/ladjs/lad) - Framework made by a former Express TC and Koa member that bundles web, API, job, and proxy servers.
 - [Ts.ED](https://github.com/tsedio/tsed) - Intituive TypeScript framework for building server-side apps on top of Express.js or Koa.js. 
+- [@egomobile/http-server](https://github.com/egomobile/node-http-server) - Very fast HTTP server with similar API to [Express](https://github.com/expressjs/express), TypeScript opimized and built-in [controller framework](https://github.com/egomobile/node-http-server/wiki/Controllers), [GraphQL](https://github.com/egomobile/node-http-server/wiki/GraphQL) and OpenAPI support out of the box.
 
 ### Documentation
 


sindresorhus:
> The submitted project should be more than 30 days old and the repo should have at least 60 stars. - https://github.com/sindresorhus/awesome-nodejs/blob/main/contributing.md

PR #1192, opened 2022-08-29
Add Async Library

sublimemm:
https://github.com/caolan/async

a 50 million download weekly library. This is the goto library for organizing async work within an application. 

 I see highland listed (which is only 60k weekly downloads) which is by the same author, and may be a spiritual successor to Async, but the API is much more complicated  and has increased scope (and is also not widely use).

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
That package used to be great, but it didn't keep up with async/await, so it's not something I would recommend these days (Yes, it does have some async/await support, but most of the docs show off callback interface).

sublimemm:
I'm not really following, there's full support for async and await since ES2017? It's mentioned right away on the Getting Started page. 

"Async accepts async functions wherever we accept a Node-style callback function"

PR #1200, opened 2022-11-29
Added Article on Password hashing in NodeJS

pareek-narendra:
We can not save the plain text as a password, and should be hashed. This article will explain everything about passwords and hashing.

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆


sindresorhus:
Thanks for the suggestion. I'm going to pass on this as the article is not high enough quality to include in this list and I also don't like including articles that seem to only be written for SEO purposes.

PR #1206, opened 2023-01-04
Add n2words

forzagreen:
[n2words](https://github.com/forzagreen/n2words) converts numbers to a human readable string. As of today, it supports 26 languages.

Examples:
```javascript
n2words(123)   // 'one hundred and twenty-three'
n2words(-1.5)  // 'minus one point five'

n2words(123, {lang: 'fr'})  // 'cent vingt-trois'
n2words(123, {lang: 'es'})  // 'ciento veintitrés'
n2words(123, {lang: 'ar'})  // 'مائة و ثلاثة و عشرون'
```
I put it under the _Humanize_ section. I could also go under _Natural language processing_.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -618,6 +618,7 @@
 - [ms](https://github.com/vercel/ms) - Tiny millisecond conversion utility.
 - [pretty-error](https://github.com/AriaMinaei/pretty-error) - Errors with less clutter.
 - [read-art](https://github.com/Tjatse/node-readability) - Extract readable content from any page.
+- [n2words](https://github.com/forzagreen/n2words) - Convert numerical numbers to written numbers, in 25+ languages: `123` → `one hundred and twenty-three`.
 
 ### Compression
 
--- a/readme.md
+++ b/readme.md
@@ -5,7 +5,16 @@
 	</div>
 	<br>
 	<br>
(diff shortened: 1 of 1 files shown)


dummib-3jYtco-qizgaw (approved):

romankurnovskii:
What is the advantage over https://www.npmjs.com/package/to-words ?

forzagreen:
@romankurnovskii n2words supports much more languages: 27 vs 9.

romankurnovskii:
> 
I see 18, so this is a question, if the issue is only in amount of languages why not to add translation or there are some more features? I use this maintained package, just want to know if there is a good reason to look at something new

forzagreen:
@romankurnovskii  
- No, there are only 9 languages. The 18 are "locales". For example, there are 10 locales for english (en-US, en-GB,...).
- Adding a new language is not just a matter of "translation". Each language has its own algorithm.
- n2words has much larger language coverage, i.e. it has major spoken languages that to-words doesn't have: Spanish, Arabic, Chinese, Russian, German...
- to-words seems to be adapted more for currencies.

romankurnovskii:
> [forzagreen](/forzagreen)

thank you @forzagreen for explanation

sindresorhus:
The package looks fine, but I don't think it's a common need. I try to curate the most common things users will need here. For this package, the user can easily search for it if they need it.

PR #1207, opened 2023-01-14
adding kloth to madscience

AlaShibanAtKlo:
Link to project: https://github.com/klothoplatform/klotho

Why it should be added: As far as we know there's no project like it. It introduces JSDoc like annotations that developers add, and klotho reads the application AST, including the annotations, and creates a deployable cloud native application. The node implementation is the most advanced and complete one.


sindresorhus:
Thanks for the suggestion. I'm unfortunately going to pass on this one.

PR #1238, opened 2023-03-30
Add Neuledge under ORMs

moshest:

⬆⬆⬆⬆⬆⬆⬆⬆⬆⬆

Check the repo here: https://github.com/neuledge/engine-js

This schema engine is unique as it allows users to define multiple states for the same entity and handle different fields and validation for each state. For example a "DraftPost" will have different fields and requirements than "PublishedPost" but both will stored and will be queried by the user in the SAME table.

```mermaid
flowchart LR
  DraftPost -->|publish| PublishedPost
  DraftPost -->|delete| Void
  PublishedPost -->|archive| ArchivedPost
```

On the repo you can find the main features section with many advantages over existing tools including Prisma.


sindresorhus:
Seems a bit premature to submit this now. I recommend waiting until the product is at 1.0.0.

